PR #63, opened 2014-03-29
Remember the scrollbar position when clicking a toolbar link

bradyvercher:
If the scroll position is above the Query Monitor panel when a toolbar link is clicked, it will be saved so that clicking the same link will return to the original scroll position. Clicking the overview link while viewing the QM panel also returns to the previous scroll position.

This makes it easier to quickly view a notice and return to the previous position without having to scroll back up a couple thousand pixels if the notice can be ignored.

Thanks for the great plugin!

### DIFF
--- a/assets/query-monitor.js
+++ b/assets/query-monitor.js
@@ -56,7 +56,8 @@ jQuery( function($) {
 	if ( !window.qm )
 		return;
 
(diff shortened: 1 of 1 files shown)


johnbillion:
I tried this on and off for a good while and never really felt comfortable with it. I'm happy with QM's scroll behaviour as it is.

PR #75, opened 2014-09-05
Added support for Google App Engine, Fixes #73

### LINKED ISSUE #73: Use of glob()  in query-monior.php on Google App Engine
First of all, thank you for this super handy plugin!!

I cam across an issue when deploying to Google App Engine:

```
12:34:23.094 PHP Warning:  Invalid argument supplied for foreach() in
/.../src/wp-content/plugins/query-monitor/query-monitor.php on line 61

12:34:23.206 PHP Warning:  Invalid argument supplied for foreach() in
/.../src/wp-content/plugins/query-monitor/query-monitor.php on line 72
```

I replaced the foreach/include lines with explicit includes for each of the files and then it worked.

aidanlane:
Using DirectoryIterator rather than glob in order to support Google App
Engine for PHP (tested on 1.9.10 - 2014-08-28).

### DIFF
--- a/dispatchers/Headers.php
+++ b/dispatchers/Headers.php
@@ -39,8 +39,11 @@ class QM_Dispatcher_Headers extends QM_Dispatcher {
 
 		require_once $this->qm->plugin_path( 'output/Headers.php' );
 
-		foreach ( glob( $this->qm->plugin_path( 'output/headers/*.php' ) ) as $output ) {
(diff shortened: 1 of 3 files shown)


aidanlane:
Oops, sorry for #93!
Thanks for all your hard work on this plugin @johnbillion, you're a champion!

PR #87, opened 2014-11-03
Add word wrap so that long strings don't break out of container

pdewouters:
Before:
screenshot 2014-11-03 09 40 37

After:

screenshot 2014-11-03 09 40 48


johnbillion:
I think this might have been made redundant by other tweaks I've made to QM's CSS in 248b482.

PR #88, opened 2014-11-26
Make the menu item toggle the view

jezzdk:
Show/hide the monitor as you wish.


johnbillion:
Thanks for the PR Jess!

I tried this out for a while but I've decided I'm not going to merge this because I often click the QM menu in order to get back to the top of QM's output. I've found that I almost never want to close the QM output once I've opened it.

Issue #82, opened 2014-10-08
Support the new admin menu behaviour in WordPress 4.1


johnbillion:
The admin menu in WordPress 4.1 will get pinned to the bottom of the screen when scrolling, if it's taller than the screen. QM's output moves underneath this as the previous behaviour was to scroll the whole menu with the screen.

Just needs an adjustment to the left hand margin in this situation.

johnbillion:
Fixed in 163783b202d9f8109208f69fe18447d5e3dadebf

PR #91, opened 2015-01-13
Prevent QM tables from hiding behind the admin menu

tnorthcutt:
Currently the admin menu overlaps the query monitor tables; this edit
adds a left margin to the outer QM div to prevent that behavior.

Example of current behavior:
2015-01-13 at 11 55 am

### DIFF
--- a/assets/query-monitor.css
+++ b/assets/query-monitor.css
@@ -138,7 +138,7 @@ GNU General Public License for more details.
 }
 
 body.wp-admin #qm {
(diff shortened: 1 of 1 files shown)


johnbillion:
Thanks. This is new behaviour in WP 4.1. See #82.

johnbillion:
Fixed in 163783b202d9f8109208f69fe18447d5e3dadebf

PR #106, opened 2015-03-13
Fixing an error for VIP Quickstart: query monitor throws an error when a...

cohix:
Fixing an error for VIP Quickstart: query monitor throws an error when a user with 2fa enabled tries to link their account without an application-specific passowrd


johnbillion:
Thanks!

cohix:
Any time!

Issue #109, opened 2015-03-19
Fatal error: Cannot redeclare class Debug_Bar


OriginalEXE:
It seems there is an issue with activating a Debug Bar plugin after Query Monitor is already activated.

I have both plugins updated to their latest version, and after I deactivate both, and activate Query Monitor first, and then Debug Bar, activation of Debug Bar fails and I get the error:
`Fatal error: Cannot redeclare class Debug_Bar in /var/www/originalexe/wp-content/plugins/debug-bar/debug-bar.php on line 20`

If I activate Debug Bar first, and then Query Monitor, there's no issue whatsoever.

P.S. I'm of course running the latest WP version.

szepeviktor:
QM has built-in Debug Bar add-on support that is what its class is already declared.
You may send a PR
http://php.net/manual/en/function.class-exists.php

johnbillion:
The problem here is that QM has to define the `Debug_Bar` class because there are several Debug Bar add-on plugins which check for the existence of this class in order to determine if they should load their panel or not.

For now, I'm going to close this ticket as a wontfix because it's not something that I can really fix, but if someone else wants to tackle it then feel free to open a PR.

PR #112, opened 2015-03-23
Fix fatal error when activating Debug Bar plugin after Query Monitor has...

OriginalEXE:
... already been activated

It's a dirty fix but it seems to be working. I tested it with multiple scenarios and could not break it.

This is related to #109


OriginalEXE:
Closing down, will open another one

PR #113, opened 2015-03-23
Fix fatal error when activating Debug Bar plugin after Query Monitor has...

OriginalEXE:
... already been activated

Related to #109 

Created a new pull request, the last one did not fix an error in case bulk activate was being used.

Anyway, hacky hack, but it works, worth considering.

### DIFF
--- a/collectors/debug_bar.php
+++ b/collectors/debug_bar.php
@@ -54,7 +54,7 @@ function register_qm_collectors_debug_bar() {
 
 	global $debug_bar;
 
-	if ( class_exists( 'Debug_Bar' ) ) {
+	if ( class_exists( 'Debug_Bar' ) || qm_debug_bar_being_activated() ) {
 		return;
 	}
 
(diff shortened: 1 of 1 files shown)


szepeviktor:
You may consider doing it in a more WP-like way.

szepeviktor:
E.g. http://codex.wordpress.org/Plugin_API/Action_Reference/load-%28page%29

OriginalEXE:
Don't think we can use that hook as it's fired after init.

I'm afraid there is no less hackier way to do this without modifying the code of Debug Bar, and that plugin is unfortunately not on GitHub.

johnbillion:
Thanks Ante, I think this is probably the right approach. I'll take a proper look in the next couple of days.

szepeviktor:
There are some actions before `init`.
http://codex.wordpress.org/Plugin_API/Action_Reference#Actions_Run_During_an_Admin_Page_Request
For example `plugins_loaded`.
Is it appropriate?

szepeviktor:
You can use `$called_script = isset( $_SERVER['SCRIPT_FILENAME'] ) ? basename( $_SERVER['SCRIPT_FILENAME'] ) : '';`

OriginalEXE:
There are multiple ones:
* muplugins_loaded
* registered_taxonomy
* registered_post_type
* plugins_loaded
* sanitize_comment_cookies
* setup_theme
* unload_textdomain
* load_textdomain
* after_setup_theme
* auth_cookie_valid
* set_current_user
But what would you do on that action, I'm not sure I understand.

What's wrong with my proposed implementation and how is checking $_SERVER global a more WP way? :)

szepeviktor:
OK-OK.
Maybe

```php
/**
 * Fires before a plugin is activated.
 *
 * If a plugin is silently activated (such as during an update),
 * this hook does not fire.
 *
 * @since 2.9.0
 *
 * @param string $plugin       Plugin path to main plugin file with plugin data.
 * @param bool   $network_wide Whether to enable the plugin for all sites in the network
 *                             or just the current site. Multisite only. Default is false.
 */
do_action( 'activate_plugin', $plugin, $network_wide );
```

OriginalEXE:
That's the first hook I checked, but it's also fired after init, so can't be used.

johnbillion:
Thanks Ante. This is in QM 2.7.4.

PR #119, opened 2015-04-08
Calling static method on class name, not variable.

### LINKED ISSUE #118: Parse error: syntax error, unexpected T_PAAMAYIM_NEKUDOTAYIM
Hi, after latest Query Monitor upgrade there was following error during attempt for activation:

`Parse error: syntax error, unexpected T_PAAMAYIM_NEKUDOTAYIM in .../public_html/www/wp-content/plugins/query-monitor/output/headers/php_errors.php on line 69`

Plugin does not work correctly now...

I am not sure what is wrong, maybe old PHP version?

ericnicolaas:
Resolves #118. Calling a static method on a variable causes a fatal error in PHP 5.2.


johnbillion:
Thanks Eric!

PR #120, opened 2015-04-16
Support for invokable objects in stack trace

Giuseppe-Mazzapica:
At the moment, `QM_Util::populate_callback()` method only supports these kind of callbacks:

 - array (both static and dymamic methods)
 - closures
 - functions

However is possible that an invokable object (an object that has an `__invoke()` method) is used as callback, in which case `populate_callback()` thrown a recoverable fatal error trying to use that invokable object as a string.

This PR handle the usage of invokable objects as callback.

### DIFF
--- a/classes/Util.php
+++ b/classes/Util.php
@@ -162,16 +162,16 @@ class QM_Util {
 
 			} else if ( is_object( $callback['function'] ) && is_callable( $callback['function'] ) ) {
 
-                if( is_a( $callback['function'], 'Closure' ) ) {
-                    $ref  = new ReflectionFunction( $callback['function'] );
-                    $file = trim( QM_Util::standard_dir( $ref->getFileName(), '' ), '/' );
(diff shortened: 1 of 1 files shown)


johnbillion:
Thank you, Giuseppe.

I've merged this into [the `develop-invokable` branch](https://github.com/johnbillion/query-monitor/commits/develop-invokable) and made some tweaks. You should be able to remove the `__toString()` method from your `WCM\WPStarter\MuLoader\MuLoader ` class with this change.

I've also done what I'm well overdue in doing, and introduced a bunch of unit tests for this.

Giuseppe-Mazzapica:
Thanks John.

Actually the `__toString()` method in `WCM\WPStarter\MuLoader\MuLoader` is not yet added, but was planned for v2.1.

If you release 2.7.4 before I release WP Starter 2.1 I'll not add it at all ;)

Issue #124, opened 2015-06-15
Localisation debugging


johnbillion:
Debugging sites that are loading in translations can be a faff, especially if things aren't working.

Brain dump of stuff I'd like to see:
- [x] Current locale
- [x] List of loaded MO files
- [ ] Available languages (locales)
- [x] Plugins / themes which don't have a translation for the current (or a given) language
- [ ] Strings on the page which don't have a translation in the current locale

pdewouters:
I've used this in the past, might get some ideas from it? https://marketpress.com/product/debug-translations/

johnbillion:
Thanks! That could be handy.

johnbillion:
Here's what this panel now looks like in QM 2.8.0:

<img width="896" alt="" src="https://cloud.githubusercontent.com/assets/208434/10569689/ec990a78-7621-11e5-941f-5f1158c37720.png">

/cc @MPolleke

PR #127, opened 2015-07-20
Add collector for Language Debugging

MPolleke:
New collector for Query Monitor, which shows Language information.

Current features:
* Mention current locale
* List Text Domain of each MO file
* List of MO files that have been asked to load, together with the file of the function call
* List result of Found / Not Found for that MO file, together with size of the file

Not implemented:
* Available languages (locales)
* List of Plugins / themes which do not try to load a MO file
* Strings on the page which don't have a translation in the current locale

It is partly based on other Collectors, and partly on the GPL v2 plugin from:
https://marketpress.com/product/debug-translations/

This is the first time I work with the code in Query Monitor, so there are probably lots of details you might want different :). If you want it reworked just let me know.
I do think this addon would be of great benefit to developers and translators already.

### DIFF
--- a/collectors/languages.php
+++ b/collectors/languages.php
@@ -1,0 +1,70 @@
+<?php
+/*
+Copyright 2009-2015 John Blackbourn
+
+This program is free software; you can redistribute it and/or modify
+it under the terms of the GNU General Public License as published by
+the Free Software Foundation; either version 2 of the License, or
+(at your option) any later version.
+
+This program is distributed in the hope that it will be useful,
+but WITHOUT ANY WARRANTY; without even the implied warranty of
+MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+GNU General Public License for more details.
+
+*/
+
+class QM_Collector_Languages extends QM_Collector {
+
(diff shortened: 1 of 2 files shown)


MPolleke:
Current layout of an admin page:
query_monitor-language_debugging

johnbillion:
This sounds great! I'm super busy for the next few days, I'll let you know when I get a chance to take a proper look :+1:

johnbillion:
See also #124

MPolleke:
Thanks.
I just forgot one thing that might be really needed. The log_file_load function in the collector was taken from the MarketPress plugin. I did not put a Copyright notice in that file though.
It should be like this:
Copyright 2013 MarketPress

johnbillion:
This has already come in handy for two separate projects. I'll get this into the next version of QM. Many thanks!

MPolleke:
No problem, I can wait :). And you're welcome. 
2 things I noticed, it uses the class="nowrap" but probably "wrap" is better for it.
Also, when using load_textdomain() directly, not all the fields are there in the array and it fails. There should be some defensive checks done if the array is there and if the fields are set.

PR #129, opened 2015-07-26
Update theme.php

robwilde:
began seeing PHP Warning:
*PHP Warning:  ltrim() expects parameter 1 to be string, array given in /var/www/hbq-tricare.dev/htdocs/wp-content/plugins/query-monitor/collectors/theme.php on line 60*

global $template does not seem to be returning the correct output and I cant find it in the Globals list so I replaced with current method to retrieve the current template page. This appears to return the intended results.

### DIFF
--- a/collectors/theme.php
+++ b/collectors/theme.php
@@ -34,7 +34,8 @@ class QM_Collector_Theme extends QM_Collector {
 
 	public function process() {
 
-		global $template;
+		// global $template;
+		$template  = get_page_template();
 
 		$template_path        = QM_Util::standard_dir( $template );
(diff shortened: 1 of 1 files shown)


johnbillion:
Thanks for the report, Rob!

I suspect that something on your site is overriding the `$template` global, which is causing this error. The variable is defined [at the bottom of template-loader.php](https://core.trac.wordpress.org/browser/tags/4.2.4/src/wp-includes/template-loader.php#L73) but it's not a "proper" global (it just happens to be global).

Using `get_page_template()` works for individual pages which are using a specific template, but doesn't work anywhere else.

I'm going to switch this over to using the `template_include` filter instead.

Issue #128, opened 2015-07-21
Stack trace tests are failing on HHVM


johnbillion:
See https://travis-ci.org/johnbillion/query-monitor/jobs/71780979#L148

> Notice: Undefined index: line in ~/tests/phpunit/test-stack-traces.php on line 124

Looks like reflection on a lambda function on HHVM doesn't behave as expected.

johnbillion:
This is caused by an HHVM bug: https://github.com/facebook/hhvm/issues/5807

Still needs a workaround in QM though.

PR #131, opened 2015-08-04
Accessibility improvements

GaryJones:
See individual commits, as they explain the accessibility improvements contained in this PR.

### DIFF
--- a/assets/query-monitor.css
+++ b/assets/query-monitor.css
@@ -316,12 +316,13 @@ body.wp-admin.folded #qm {
 
 .qm span.qm-true,
 .qm td.qm-true {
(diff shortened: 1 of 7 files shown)


johnbillion:
Neato! I'll have a check through this shortly.

GaryJones:
FWIW, I have absolutely no idea what the cause of those HHVM failures are.

Do you not allow HHVM to fail?

johnbillion:
See #128 for the HHVM failures. I'm not allowing tests to fail on HHVM currently because ideally they shouldn't.

johnbillion:
These changes, along with other improvements, went into 2.12.0.

PR #157, opened 2015-12-10
Remove is_comments_popup() from query_monitor_conditionals

westonruter:
It is deprecated as of https://core.trac.wordpress.org/changeset/35848

The internet broke. (Deprecation notices started appearing at `shutdown` in Query Monitor and corrupted JSON Ajax responses.)


westonruter:
In the mean time, a workaround is:

```php
add_filter( 'query_monitor_conditionals', function ( $conds ) {
    return array_diff( $conds, array( 'is_comments_popup' ) );
} );
```

westonruter:
Looks like Travis is broken, but not because of my PR :smile:

PR #160, opened 2015-12-13
Remove Deprecated `'is_comments_popup` when WP 4.5

mikeschinkel:
Using bleeding edge 4.5 I am getting an error from Query Monitor because `is_comments_popup()` is now deprecated in WP 4.5.  

I've added a quick fix, which is similar to your other filter on likes 98-103 so merge if you like the fix or consider fixing another way if you. Mostly just wanted you to be aware.


johnbillion:
#157

johnbillion:
I'm just going to remove this, by the way. Not bothered about back-compat. :+1:

mikeschinkel:
Cool. 

RE: #157, I searched for `is_comments_popup` on WP.org, but didn't search here. Doh! 

Sorry for the dup. :)

PR #162, opened 2015-12-22
Allow clicking URL's to open

hubdotcom:
Chrome by default if copying the URL converts the `<br>` into `%20`, resulting in malformed requests, this would allow 1 click access to the URL's

### DIFF
--- a/output/Html.php
+++ b/output/Html.php
@@ -154,7 +154,8 @@ abstract class QM_Output_Html extends QM_Output {
 	 * @return string      The URL formatted with markup.
 	 */
 	public static function format_url( $url ) {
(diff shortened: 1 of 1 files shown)


johnbillion:
Thanks for the PR. What platform are you on? In Chrome on OS X, copying the URL works as expected. The `<br>`s get removed when the URL is pasted into the address bar.

hubdotcom:
It's now working on OS X - Chome - guess they patched this. The PR would still allow 1 click access to those URL's via the a tag

johnbillion:
Thanks for the PR, hubdotcom, but I've decided that linking these URLs isn't such good UX. I have some other plans for improving this section in the future.

Issue #170, opened 2016-03-08
Send no-cache headers when the Query Monitor cookie is set


simonwheatley:
Query Monitor optionally sets a cookie which allows viewing Query Monitor information on requests where the user is not logged in. `no-cache` headers should be set on these requests, so the responses (with QM information) don't get cached, e.g. in Varnish.

johnbillion:
Fixed in #171.

PR #171, opened 2016-03-08
Send nocache headers when QM is active

nickdaugherty:
This helps prevent caching of QM data, especially when using the
authentication cookie on anonymous requests (which may otherwise cache)


simonwheatley:
See https://github.com/johnbillion/query-monitor/issues/170

PR #176, opened 2016-04-06
Add support for the AMP plugin

johnbillion:
The AMP plugin by Automattic short circuits template output and therefore doesn't trigger the necessary hooks that QM uses to determine that a page has indeed been output as expected (this is intentional on the part of QM, so that it doesn't display its output when a plugin is generating, for example, an XML feed in an action on the `template_redirect` hook).

This adds support for this plugin so QM's output appears at the bottom of AMP output.

### DIFF
--- a/dispatchers/Html.php
+++ b/dispatchers/Html.php
@@ -32,6 +32,7 @@ class QM_Dispatcher_Html extends QM_Dispatcher {
 		add_action( 'admin_footer',               array( $this, 'action_footer' ) );
 		add_action( 'login_footer',               array( $this, 'action_footer' ) );
 		add_action( 'embed_footer',               array( $this, 'action_footer' ) );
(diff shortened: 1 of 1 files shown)


crstauf:
while working on adding AMP capability to my own site, I stumbled on the stubbornness of `ensure_ajaxurl`. Adding the following code inside the function before printing (seems to) resolve it's stubbornness when not needed:

`if (!current_user_can('view_query_monitor') && !QM_Dispatcher::user_verified()) return;`

I also have this code before in my template for handling AMP requests:

```if (!current_user_can('view_query_monitor') && !QM_Dispatcher::user_verified())
	add_filter('qm/dispatch/html','__return_false',999999999);```

I'm just starting on experimenting with having AMP work on my site, so this stuff may not entirely sufficient or needed, but I thought what I've got thus far may be helpful.

PR #178, opened 2016-04-13
#174 - Export as html file

Mte90:
schermata 2016-04-13 alle 21 04 10

This pr add an export button that generate on the file a html file with css and js from query monitor.
The css is in bundle with the html and support no-js, in the case there is connection load the js (jquery from cdnjs) from the site that generated that file.

Plus to improve that I am thinking to print near the 'Query Monitor' label the url of the page so in the file is possible to check what is the url that generated that, but I prefer a feedback for that.

PR #179, opened 2016-04-13
Scroll to all the query monitor area

Mte90:
With https://github.com/johnbillion/query-monitor/pull/178 we have a button that when you click on the admin bar is covered from the browser window.
With that the scroll position on click on the admin bar move the visibile area also to the query monitor title and botton.


Mte90:
Sorry wrong branch for the pr, I've opened a new pr with the correct branch

PR #187, opened 2016-04-27
Add option for showing all hooks

misenhower:
Currently, Query Monitor's Hooks section only displays hooks that have been called at least once by `do_action()`. Filters generally don't show up in Query Monitor, however, since `apply_filters()` doesn't increment the `$wp_actions` counter.

This change adds an option to display all registered hooks instead of just the ones called with `do_action()`. This is useful for determining which filters may have been called (as well as simply listing all hooks that are registered in the installation for debugging/diagnostic purposes).

With this change, defining `QM_SHOW_ALL_HOOKS` as true will make all hooks appear.


Mte90:
pretty cool!

PR #193, opened 2016-06-04
fixed Undefined index: cache_misses

toddlahman:
Fixed the Undefined index: cache_misses notice, which surfaced on PHP 7.0.7, and loaded a screen full of notices and warnings with wp_debug set to true. The headers already sent warnings remain. Below is a list of notices and warnings on plugin activation.

Notice: Undefined index: cache_misses in /var/www/html/hci/wp-content/plugins/query-monitor/collectors/cache.php on line 51

Warning: Cannot modify header information - headers already sent by (output started at /var/www/html/hci/wp-content/plugins/query-monitor/collectors/cache.php:51) in /var/www/html/hci/wp-content/plugins/query-monitor/output/Headers.php on line 25

Warning: Cannot modify header information - headers already sent by (output started at /var/www/html/hci/wp-content/plugins/query-monitor/collectors/cache.php:51) in /var/www/html/hci/wp-content/plugins/query-monitor/output/Headers.php on line 25

Warning: Cannot modify header information - headers already sent by (output started at /var/www/html/hci/wp-content/plugins/query-monitor/collectors/cache.php:51) in /var/www/html/hci/wp-content/plugins/query-monitor/output/Headers.php on line 25

Warning: Cannot modify header information - headers already sent by (output started at /var/www/html/hci/wp-content/plugins/query-monitor/collectors/cache.php:51) in /var/www/html/hci/wp-content/plugins/query-monitor/output/Headers.php on line 25

Warning: Cannot modify header information - headers already sent by (output started at /var/www/html/hci/wp-content/plugins/query-monitor/collectors/cache.php:51) in /var/www/html/hci/wp-content/plugins/query-monitor/output/Headers.php on line 25

Warning: Cannot modify header information - headers already sent by (output started at /var/www/html/hci/wp-content/plugins/query-monitor/collectors/cache.php:51) in /var/www/html/hci/wp-content/plugins/query-monitor/output/Headers.php on line 25

Warning: Cannot modify header information - headers already sent by (output started at /var/www/html/hci/wp-content/plugins/query-monitor/collectors/cache.php:51) in /var/www/html/hci/wp-content/plugins/query-monitor/output/Headers.php on line 25

Warning: Cannot modify header information - headers already sent by (output started at /var/www/html/hci/wp-content/plugins/query-monitor/collectors/cache.php:51) in /var/www/html/hci/wp-content/plugins/query-monitor/output/Headers.php on line 25

[...]

PR #194, opened 2016-06-06
fixed PHP notice in QM_Collector_Cache

petenelson:
Was getting a PHP notice on the `cache_misses` array index after updating to 2.11.0

```
Notice:  Undefined index: cache_misses in /srv/www/wordpress-default/htdocs/wp-content/plugins/query-monitor/collectors/cache.php on line 51
```


johnbillion:
Thanks Pete, someone beat you to this in #193 :)

PR #202, opened 2016-06-26
allow query monitor to load when using wp-cli

### LINKED ISSUE #142: Activation with WP-CLI fails to set up view_query_monitor capability
[As reported in the support forums](https://wordpress.org/support/topic/wp-cli-activation-doesnt-seem-to-attach-the-admin-menu).

Activating QM with WP-CLI does not result in the `view_query_monitor` capability being assigned to admins.

Either WP-CLI isn't triggering the activation hook, or QM's activation hook isn't compatible with what happens when WP-CLI does fire it.

phh:
Fixes #142

### DIFF
--- a/query-monitor.php
+++ b/query-monitor.php
@@ -30,7 +30,7 @@ if ( defined( 'QM_DISABLED' ) and QM_DISABLED ) {
 	return;
 }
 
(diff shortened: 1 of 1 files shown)


johnbillion:
Thanks for this, @phh, but this change defeats most of the point of this line of code which is to avoid loading QM when using WP-CLI.

I have a better fix for this problem coming up so I'll close this PR.

PR #204, opened 2016-07-11
add alert color for failed resource and HTTP requests

crstauf:
In my development workflow, I will often see QM in the admin bar with a red background, and immediately investigate to see what PHP error is in my code, but most of the time, it's just a script or stylesheet that failed to load, or an HTTP request that failed.

This may just be a personal preference, but I'd like to prioritize PHP warnings by creating another error type/color for those events, to distinguish between a coding error and the instability of the Internet.

I wasn't sure about creating a whole new error type, so the attached code merely adds and uses a new class, `qm-alert`.

### DIFF
--- a/assets/query-monitor.css
+++ b/assets/query-monitor.css
@@ -69,7 +69,7 @@ GNU General Public License for more details.
 }
 
 #wpadminbar .qm-alert {
-	background-color: #f90 !important;
+	background-color: #f70 !important;
 }
 
 #wpadminbar .qm-error > a,
--- a/assets/query-monitor.css
+++ b/assets/query-monitor.css
@@ -68,7 +68,12 @@ GNU General Public License for more details.
 	background-color: #c00 !important;
 }
 
-#wpadminbar .qm-error > a {
+#wpadminbar .qm-alert {
+	background-color: #c75411 !important;
+}
+
+#wpadminbar .qm-error > a,
+#wpadminbar .qm-alert > a {
 	color: #fff !important;
 }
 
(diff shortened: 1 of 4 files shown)


johnbillion:
Thanks, I like this idea.

The colour needs tweaking (I actually prefer the dark orange colour in your first commit but it's too close to red). `#f70` seems to work well.

johnbillion:
Screenshot of `#f70`:

screen shot 2016-07-12 at 14 50 34

crstauf:
would it be worth it to create a new error type, or do you think using a new class name is sufficient?

johnbillion:
I think it's fine just to adjust the CSS output. Each collector still treats its own errors as errors, just the visual output in the admin toolbar can differ. 👍

crstauf:
the main reason I'm curious about implementing a new error type is that [Query Monitor Extend](https://github.com/crstauf/query-monitor-extend) bases it's gradient creation on the number of each error type reported by the collector. not creating a separate error type will render that functionality useless for including failed requests and asset calls.

crstauf:
example:
screen shot 2016-07-12 at 10 25 02 am
screen shot 2016-07-12 at 10 25 22 am

johnbillion:
Ah interesting.

In that case, let's switch the `HTTP Requests` component to use a new error type called `alert`. I think the `Scripts & Styles` component should continue using errors that appear in bright red, as script dependency errors are fully fledged errors in my opinion.

crstauf:
almost there...
screen shot 2016-07-12 at 10 55 36 am

crstauf:
committed 503ec2beac1fc271ad748a32b94040dbd9ea2ae3 on [Query Monitor Extend](https://github.com/crstauf/query-monitor-extend) to support new color and error type

johnbillion:
Thanks @crstauf 👍

PR #214, opened 2016-07-25
Enhancements for Filtering Table Data

crstauf:


johnbillion:
Similarly to #223, I'm going to close this off as the table filtering will be completely reworked as part of #483 .

Cheers!

PR #219, opened 2016-07-30
set 'data-qm-sort-weight' attribute on cells with standard decimal value

crstauf:
No matter what I do, I cannot seem to get [`dataset.qmSortWeight`](https://github.com/johnbillion/query-monitor/blob/master/assets/query-monitor.js#L336) to be used for the sorting value... switching to `getAttribute('data-qm-sort-value')` works perfectly in my tests. Searching QM directory for 'weight' results in the one script line, and `font-weight` declarations in the stylesheet; if I've missed a table that uses attributes for the sorting value, @johnbillion, please let me know.

### DIFF
--- a/output/html/db_callers.php
+++ b/output/html/db_callers.php
@@ -74,7 +74,7 @@ class QM_Output_Html_DB_Callers extends QM_Output_Html {
 					}
 				}
 
-				echo '<td class="qm-num">' . esc_html( $stime ) . '</td>';
+				echo '<td class="qm-num" data-qm-sort-weight="' . esc_attr( number_format( $row['ltime'], 4, '.', '' ) ) . '">' . esc_html( $stime ) . '</td>';
 				echo '</tr>';
 
 			}
--- a/assets/query-monitor.js
+++ b/assets/query-monitor.js
@@ -335,10 +335,10 @@ jQuery( function($) {
 				
 				for (var k = 0, l = row_width; k < l; k++) {
 					var e = columns[i+k];
-					
-					var data = e.dataset.qmSortWeight;
-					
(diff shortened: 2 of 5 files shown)


crstauf:
this PR would also present a solution for https://wordpress.org/support/topic/does-sort-by-query-time-work:

```
echo '<td class="qm-num qm-row-time' . esc_attr( $td ) . '" data-qm-sort-value="' . esc_attr( number_format( $row['ltime'], 4 ) ) . '">'. esc_html( $stime ) . "</td>\n";
```

crstauf:
hummm. well it embarrasses me to say, but apparently `dataset.qmSortWeight` translates into an attribute named `data-qm-sort-weight`; did not know that. I'll change this PR then to add attributes to cells with decimal values.

johnbillion:
> apparently dataset.qmSortWeight translates into an attribute named data-qm-sort-weight; did not know that.

Same here! I didn't know that either.

output/html/db_callers.php
```diff
@@ -74,7 +74,7 @@ public function output() {
 					}
 				}
 
-				echo '<td class="qm-num">' . esc_html( $stime ) . '</td>';
+				echo '<td class="qm-num" data-qm-sort-weight="' . esc_attr( number_format( $row['ltime'], 4, '.', '' ) ) . '">' . esc_html( $stime ) . '</td>';
```
johnbillion: Seeing as `$row['ltime']` here is an actual float, there's no need to pass it through `number_format()`. It can be used as-is.

johnbillion:
Merged, thanks 👍

PR #220, opened 2016-08-01
add toggle for PHP errors 'Call Stack' column

crstauf:
I've got a PHP error with a pretty long stack, which made me wonder about having the first item in the stack, and adding a toggle to reveal the entire trace stack. Thoughts?


johnbillion:
I've thought about this before and decided against it. The reason the error call stack is shown in full is that, unlike the call stack for database queries, you're much more likely to want to see what caused the error.

crstauf:
alright, I'll implement into QMX instead.

PR #226, opened 2016-08-08
Table accessibility improvements

GaryJones:
Many of the output tables were using a single row `thead` as a way to _title_ the table. This is incorrect, and a `caption` element is both more semantic and user-friendly to screen-readers (it won't potentially get read out again and again like a table heading would).

Most of the tables also didn't have `scope` attributes for `th` elements. This commit adds explicit values in for the `scope` attributes.

All bar a couple of instances, the tables should visually look exactly the same as they do now.

There are two discrepancies:

1. The _Conditionals_ table isn't really tabular data, and may need further improvements to indicate to screen readers which conditionals are true and which are not. Something like the following may benefit screen readers:

    ```
    <h3 class="screen-reader-text">Satisfied Conditionals</h3>
    <ul>
        <li>...</li>
        <li>...</li>
    </ul>
    <ul aria-hidden="true">
        <li>...</li>
        <li>...</li>
    </ul>
    ```

    Where the second list (unsatisfied conditionals) is hidden from screen readers - just indicate success to them, like the admin toolbar dropdown does. Obviously some CSS is needed to then turn that into the appearance similar to how it currently is.

2. The _Admin_ table is sufficiently complex, that [this table validator](http://wet-boew.github.io/wet-boew-legacy/v3.1/demos/tableparser/validator-htmltable.html#fnb2) suggests using `id` and `headers` attributes instead of `scope`. As such, I've not added `scope` in to this table so it's no worse than it currently is.

I've left each output file as an individual commit to allow easier following, and notes per commit. You may wish to squash these down during merging.

### DIFF
--- a/assets/query-monitor.css
+++ b/assets/query-monitor.css
@@ -286,7 +286,8 @@ body.wp-admin.folded #qm {
 }
 
 .qm td,
-.qm th {
+.qm th,
+.qm caption {
 	background: #fff !important;
 	box-sizing: border-box !important;
 	text-align: left !important;
@@ -303,11 +304,9 @@ body.wp-admin.folded #qm {
 	-webkit-font-smoothing: auto !important;
 }
 
-.qm .qm-totally-legit-spacer td {
-	background-color: #f1f1f1 !important;
-	border-right-color: #f1f1f1 !important;
-	border-left-color: #f1f1f1 !important;
-	height: 20px !important;
+.qm caption,
+.qm th[scope=col] {
+	border-bottom-width: 0 !important;
 }
 
 .qm tbody tr:hover th,
--- a/output/html/admin.php
+++ b/output/html/admin.php
@@ -31,34 +31,30 @@ class QM_Output_Html_Admin extends QM_Output_Html {
 
 		echo '<div class="qm qm-half" id="' . esc_attr( $this->collector->id() ) . '">';
 		echo '<table cellspacing="0">';
-		echo '<thead>';
+		echo '<caption>' . esc_html( $this->collector->name() ) . '</caption>';
+		echo '<thead class="screen-reader-text">';
 		echo '<tr>';
-		echo '<th colspan="2">' . esc_html( $this->collector->name() ) . '</th>';
+		echo '<th>' . esc_html__( 'Data', 'query-monitor' ) . '</th>';
+		echo '<th>' . esc_html__( 'Property', 'query-monitor' ) . '</th>';
+		echo '<th>' . esc_html__( 'Value', 'query-monitor' ) . '</th>';
 		echo '</tr>';
 		echo '</thead>';
 		echo '<tbody>';
 
 		echo '<tr>';
-		echo '<td class="qm-ltr">get_current_screen()</td>';
-		echo '<td class="qm-has-inner">';
+		echo '<th class="qm-ltr" rowspan="' . absint( count( get_object_vars( $data['current_screen'] ) ) + 1 ) . '">get_current_screen()</th>';
+		echo '</tr>';
 
-		echo '<table class="qm-inner" cellspacing="0">';
-		echo '<tbody>';
 		foreach ( $data['current_screen'] as $key => $value ) {
 			echo '<tr>';
 			echo '<th>' . esc_html( $key ) . '</th>';
 			echo '<td>' . esc_html( $value ) . '</td>';
 			echo '</tr>';
 		}
-		echo '</tbody>';
-		echo '</table>';
-
-		echo '</td>';
-		echo '</tr>';
 
 		echo '<tr>';
-		echo '<td class="qm-ltr">$pagenow</td>';
-		echo '<td>' . esc_html( $data['pagenow'] ) . '</td>';
+		echo '<th class="qm-ltr">$pagenow</th>';
+		echo '<td colspan="2">' . esc_html( $data['pagenow'] ) . '</td>';
 		echo '</tr>';
 
 		$screens = array(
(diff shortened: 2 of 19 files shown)


johnbillion:
The a11y improvements look good, thanks.

The only issue I can see at first glance is the separation of the Scripts table from the Styles table. The purpose of the totally-legit-spacer is to provide the visual appearance of two tables, but use one table in the markup so that the column widths of both tables match.

Here's a before/after comparison:

**Before:**

before

**After:**

after

I'd prefer not to lose the matching column widths because it's a great visual enhancement over two separate tables.

GaryJones:
I hadn't spotted that, but to be honest, a slightly visual difference in tables column widths between two different types of assets is less of a concern (to me) than making the data far more accessible to everyone.

There are multiple tables with various column widths on the page, so it's not like these need to line-up with everything else.

Even though we know scripts and styles are handled as a generic Dependency, those who want to see what (un-related) scripts and styles are enqueued on a particular page, may not know or care.

There may be a way to force a certain % width on the variable size column(s), but I don't think it's warranted. If a long path needs more space, why force a break?

From an a11y point of view, two tables gives two separate (and correct) captions, and the correct scoping (automatic or explicit).

More importantly, two tables are more valid markup than the `thead` -> `tbody` -> `thead` -> `tbody` that currently exists. A quick check in a validator will show that:
```html
<!DOCTYPE HTML SYSTEM>
<html>
<head><title>title</title></head>
<body>
<table>
<thead><tr><th>1</th></tr></thead>
<tbody><tr><td>2</td></tr></tbody>
<thead><tr><th>1</th></tr></thead>
<tbody><tr><td>2</td></tr></tbody>
</table>
</body>
</html>
```
and:
```html
<!DOCTYPE HTML SYSTEM>
<html>
<head><title>title</title></head>
<body>
<table>
<thead><tr><th>1</th></tr></thead>
<thead><tr><th>1</th></tr></thead>
<tbody><tr><td>2</td></tr></tbody>
</table>
</body>
</html>
```
are both invalid, since multiple `thead` elements are not allowed in a single table. Screen readers and other assistive technologies will generally handle valid markup better than invalid markup.

I'm strongly in favour of a11y for everyone >>> visual enhancement for sighted users on this one.

crstauf:
I'm curious about adapting QM to be more accessible... given that it's primarily a debugging tool used by developers, what are the potential benefits to QM users? I certainly see the advantage in making non-breaking changes, but ones that negatively impact the output seem unnecessary. I for one appreciate the column widths remaining the same for the Assets tables; given that QM is primarily a debugging tool, while following standards is a generally smart and beneficial practice, it remains (in my opinion) unnecessary.

@GaryJones is there an acceptable markup for accessibility that would allow the two to remain as one?

GaryJones:
You think there aren't any blind developers? Of course there are. Just because it's a tool is for debugging, doesn't mean it shouldn't be accessible. 

I've already suggested that % widths could be applied to cells if you wanted visual alignment of two completely distinct tables, but I've also said why I don't think that's warranted (un-neccessary visual line-breaks of URLs).

crstauf:
@GaryJones I'm sure there are blind developers, I'm just wondering about their workflow, and if QM is part of that.

and in order for the cells to align, every column would require a specific percentage width, which is not preferred; allowing the browser to determine the table column widths is the much better option.

GaryJones:
> allowing the browser to determine the table column widths is the much better option

Exactly ;-)

GaryJones:
@amandarush - Do you use Query Monitor (or would you, if it was more accessible)?

Do you know of any other screen-reader users who use it, or would if it were more accessible?

rianrietveld:
Fresh into the discussion: I know at least 9 blind or almost blind WordPress developers, but that is not the issue here I guess. 
<lecture>Coding accessible is just good practice, it will also benefit people using other AT, speech recognition software, and who knows what other software that expects semantically correct HTML.</lecture>

Setting a width in % might be a good compromise?

crstauf:
@GaryJones @rianrietveld is there any acceptable coding practice that would allow the table to retain it's visual structure, but make the table itself more accessible?

GaryJones:
> is there any acceptable coding practice that would allow the table to retain it's visual structure, but make the table itself more accessible?

The visual styling (CSS) has to come second to the table structure (HTML) for it to be accessible, so it really comes down to what CSS or JS you know.

```
.table-fixed {
    table-layout: fixed;
    width: 100%;
}
```

That would give equal widths to all columns across a table, and if applied to both tables, they would of course then be in alignment.

Alternatively, we might need some JavaScript support:
http://stackoverflow.com/questions/456859/any-way-to-synchronize-table-column-widths-with-html-css# (a couple of similar solutions)

@johnbillion Is it worth me cherry-picking all of the other commits into a new PR, and getting that merged, and then treating this table(s) issue as a separate PR?

johnbillion:
> @johnbillion Is it worth me cherry-picking all of the other commits into a new PR, and getting that merged, and then treating this table(s) issue as a separate PR?

Yes please! 👍

GaryJones:
> Yes please! 👍

Done as #230.

joedolson:
I question the proposition that having equal width columns in these two tables is actually an improvement. To me, and this is obviously only a personal observation, this makes it more difficult to scan through and find the break point between the two groups of information. 

There may be advantages to having the columns widths the same, but I don't really see it.

GaryJones:
Is anyone going to stop using the plugin, because two distinct tables no longer visually align? Probably not.

Do these changes allow more WP developers to use this plugin in their day to day work flow? Absolutely.

johnbillion:
Merged, thanks everyone.

GaryJones:
@johnbillion So, #230 got merged? That still leaves the lack of improvements on the assets table, correct?

PR #227, opened 2016-08-09
Save user options

crstauf:
This PR adds the ability for a user to save their desired default sorting column/order, and other preferences. By holding the Shift key while clicking, the selection can be saved using AJAX and  `update_user_option`, and will be recalled when QM's output is regenerated.

So for example, if you want the Queries panel to be sorted by longest time, hold the Shift key, and click the down arrow under the Time column. That will save the configuration to `{$wpdb->prefix}_qm_prefs`, and will sort the table by that column from there on out.

One thing to note: to allow proper ordering of the DB Queries and HTTP panels by number, I added a new value to the data array, `i`.

This is probably not a finished product, but it's a large start. You can see more examples with [QMX's `qm-save-user-options` branch](https://github.com/crstauf/query-monitor-extend/tree/qm-save-user-options).

### DIFF
--- a/assets/query-monitor.css
+++ b/assets/query-monitor.css
@@ -577,6 +577,8 @@ select.qm-filter {
 	margin: 0 auto !important;
 }
 
+.qm .qm-sort:focus { box-shadow: none !important; }
+
 .qm .qm-sort-asc {
 	margin-bottom: 1px !important;
 }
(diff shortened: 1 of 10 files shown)


johnbillion:
Thanks for the PR Caleb, but I've decided that for now this isn't something I want to add to QM.

Note that in version 2.16, QM now automatically remembers your choice of filter in the browser's localStorage. The behaviour is all client side.

PR #234, opened 2016-09-09
Use latest HHVM on Trusty

ntwb:
HHVM is no longer being developed for Ubuntu Precise, to use a current version of HHVM Ubuntu Trusty is required. Also HHVM on Trusty requires MySQL 5.6.

As the build matrix cannot mix and match OS environments with MySQL versions I've only added HHVM with `nightly` WordPress version for both single and multisite.

If WordPress versions `4.5` and `latest` are required a follow up pull request can be made.

See also: https://core.trac.wordpress.org/changeset/37555/trunk/.travis.yml

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,7 +14,6 @@ php:
   - 5.6
   - 5.3
   - 5.2
-  - hhvm
 
 env:
   # WP nightly:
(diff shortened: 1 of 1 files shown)


ntwb:
Yay, the current version is HHVM 3.15.0, previously it was 3.6.6 🎉

PR #235, opened 2016-09-09
Use latest HHVM on Trusty (Alternate matrix config)

ntwb:
Second guessing myself, this is basically a duplicate of #234 

In this PR though I've tried keeping HHVM in the build matrix, let's see what happens.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,6 +15,15 @@ php:
   - 5.3
   - 5.2
   - hhvm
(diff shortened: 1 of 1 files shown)


ntwb:
Nope, not allowed, the error:

`The log message was: Build config file had a parse error: mapping values are not allowed in this context at line 18 column 9.`

Was worth a shot...

Issue #223, opened 2016-08-05
Plugin names with same beginning conflict in Component filtering


crstauf:
Perfect example is Query Monitor (`Plugin: query-monitor`) and Query Monitor Extend (`Plugin: query-monitor-extend`):

![screen shot 2016-08-04 at 9 36 08 pm](https://cloud.githubusercontent.com/assets/4573033/17423759/84e59852-5a8b-11e6-95f5-d2c6e8b83d74.png)

The simple solution is just to add a space (" ") after the component, but wondering if we should consider a less common delimiter, such as `|` or `,`?

johnbillion:
I'm going to close this as it's not something I'm interested in spending time fixing myself, but feel free to open a PR if you find a good solution.

johnbillion:
Closing this off as it's an edge case and it'll ultimately get fixed in #483 when the filtering operates at the data level instead of the display level.

PR #245, opened 2016-12-11
Fix issue 223 - Plugin names with same beginning conflict in Component filtering

crstauf:
add space to subject/component element attributes for more accurate cell highlighting/selecting; fixes issue #223


johnbillion:
Sorry this never got merged, but I'm going to close this off because as part of my work in #483 to implement client-side rendering, I'm going to be adding proper state-driven filtering of the data so I can rip out all this jQuery. That way we won't have to rely on any string matching jankyness.

Cheers!

PR #250, opened 2017-02-09
fix #241 PHP Notice when using Timber Library

### LINKED ISSUE #241: PHP Notice when using Timber Library
PHP Notice when using Timber Library:

```
Missing argument 2 for QM_Output_Html::output_filename(), 
called in wp-content\plugins\query-monitor\output\html\theme.php on line 100
```

Can be fixed by passing empty string as second argument to ```output_filename()``` method.

nagayama:
#241 `self::output_filename` requires 2 arguments. There seems no needs to use it here.


johnbillion:
Thanks! 👍

PR #255, opened 2017-06-07
fix php errors when $wp_scripts is null

cgalvarez:
While debugging WordPress with XDebug, I noticed that in the login page at `/wp-admin/` the debugger stopped due to PHP errors because of `$wp_scripts` is null and it's not checked. Don't know if it happens on any other page, but this PR fixes that.

### DIFF
--- a/collectors/assets.php
+++ b/collectors/assets.php
@@ -32,7 +32,7 @@ class QM_Collector_Assets extends QM_Collector {
 		global $wp_scripts, $wp_styles;
 
 		$this->data['header']['styles'] = $wp_styles->done;
-		$this->data['header']['scripts'] = $wp_scripts->done;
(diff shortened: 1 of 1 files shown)


cgalvarez:
Don't know why the tests failed. It seems that one of the four tests hung up.

johnbillion:
Thanks, I restarted that test job.

I'll need to look into why the `$wp_scripts` array is empty here and why I haven't seen this problem before. I won't have much time to look into this over the next few weeks, but thanks for the PR and I'll do my best to take a look as soon as I can.

johnbillion:
I'm going to close this off as it's not an error I've ever seen and I've not had any other reports about it. Feel free to reopen if you can provide steps to reliably reproduce the problem.

Cheers!

PR #264, opened 2017-08-09
Ability to silence PHP errors from select plugins

### LINKED ISSUE #239: Ability to silence PHP errors from selected plugins
When working on a site where you're not 100% in control of all the code -- for example, shared plugins on the WordPress.com VIP platform -- it can be annoying to see a persistent notification for PHP notices caused by poor code.

QM should provide a feature to mute specific error levels for specific components via a filter. For example:

``` php
add_filter( 'qm/collect/silent_php_errors', function( array $silent ) {
    $silent[ 'chartbeat' ] = E_NOTICE & E_STRICT;
    $silent[ 'theme' ]     = QM_E_DEPRECATED;
    return $silent;
} );
```

This should work similarly to the silencing of the HTTP API when Airplane Mode is active, and the silencing of `@-` suppressed PHP errors. They errors should still be listed but shouldn't trigger a notification in the admin toolbar.

dsawardekar:
I've added support for silencing PHP errors from select plugins in this PR. This fixes #239. The approach is pretty close to the filter hook mentioned on that ticket.

```php
	add_filter( 'qm/collect/silent_php_errors', function( $table ) {
		$table['bad-plugin'] = E_ALL & ~E_NOTICE & ~E_WARNING;
		return $table;
	} );
```
Here's the full [bad-plugin example](https://gist.github.com/dsawardekar/2a3e22d57d43de3ecfc726c4c28e7186). I'm excluding Notices & Warnings from it.

I also included a small filter hook(qm/collect/hide_silent_php_errors) to hide such errors from QM entirely. It is set to false by default. I found it handy in development environments.

Screenshot: http://imgur.com/a/kG70A

### DIFF
--- a/collectors/php_errors.php
+++ b/collectors/php_errors.php
@@ -182,6 +182,145 @@ class QM_Collector_PHP_Errors extends QM_Collector {
 		restore_error_handler();
 	}
 
+	/**
+	 * Runs post-processing on the collected errors and updates the
+	 * errors collected in the data->errors property.
+	 *
+	 * Any unreportable errors are placed in the data->filtered_errors
+	 * property.
+	 */
+	public function process() {
+		if ( ! empty( $this->data ) && ! empty( $this->data['errors'] ) ) {
+			/**
+			 * Filters out non-reportable errors based on the errors
+			 * table.
+			 *
+			 * The table is empty by default. Users can specify errors
+			 * to filter by overriding it and adding PHP Error flags.
+			 *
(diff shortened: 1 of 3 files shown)


johnbillion:
Thanks for the PR! I'll see what time I can find next week to look at this.

dsawardekar:
@johnbillion Any thoughts on this?

johnbillion:
Closing this in favour of #287 which is just an extension of this PR's branch with some additional tweaks. Thanks for your work on this @dsawardekar.

PR #265, opened 2017-08-21
Add constant to disable error handling

rasmusbe:
Adds support to disable the error handler in QM via constant.

If the site is using for example [Sentry](https://sentry.io) or [BugSnag](https://www.bugsnag.com/) or any other error tracker for handling errors we don't want QM overriding the error handling.


johnbillion:
Thanks for the PR. I renamed the constant to `QM_DISABLE_ERROR_HANDLER` and this has gone into QM 2.17.0.

PR #268, opened 2017-09-12
remember the picked value in all the filters

Mte90:
Helpful on screen refresh :-)


johnbillion:
I've made some tweaks to this. Let me know what you think.

johnbillion:
One task left that I'm working on: If the stored value is not present in the list of available options, it shouldn't try to select it (you end up with an empty selected item).

Mte90:
https://developer.mozilla.org/it/docs/Web/API/Window/localStorage Localstorage is supported from all the browser from years (chrome 3 and firefox 4 as example) so I don't think that check is required

PR #274, opened 2017-10-18
Log and display user capability checks

johnbillion:
This introduces a new panel which shows user capability checks that have been performed during the current page load.

The panel is quite noisy at the moment as it includes all cap checks from both the admin menu and the admin toolbar, which I think need filtering out by default.

Needs some thorough testing.

screen shot 2017-10-18 at 13 53 37

### DIFF
--- a/assets/query-monitor.css
+++ b/assets/query-monitor.css
@@ -623,6 +623,7 @@ select.qm-filter {
 }
 
 .qm-hide,
+.qm-hide-user,
 .qm-hide-result,
 .qm-hide-name,
 .qm-hide-type,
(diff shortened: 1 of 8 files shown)


Mte90:
I am trying but seems that the table width in my case is more bigger than the others tables.

johnbillion:
I've switched this panel to use the collapsed view for the caller column. Greatly reduces the space that it takes up.

screen shot 2017-10-18 at 17 07 06

johnbillion:
The cap check result and any arguments which were passed are now shown too.

If any cap check is performed against another user (eg. with `user_can( ... )`) then an extra column shows up which shows the user ID.

shadyvb:
Can't thank you enough for this, been doing this manually, and painfully, for so long!

Issue #278, opened 2017-11-01
Implement ARIA state on toggle buttons


johnbillion:
The buttons for toggling open stack traces and other information should implement the correct ARIA attributes and state so they are more accessible.

I'm not sure on specifics, but here's some starting info: https://www.w3.org/WAI/GL/wiki/Using_WAI-ARIA_state_and_property_attributes_to_expose_the_state_of_a_user_interface_component#Example_1:_A_toggle_button

rianrietveld:
Commented with #285: this needs `aria-expanded`, not `aria-pressed`.

ivdimova:
I assume we can close this one @johnbillion ?

johnbillion:
Fixed in #285

PR #285, opened 2017-11-24
Adding ARIA attributes to toggle buttons

ivdimova:
For #278

### DIFF
--- a/output/Html.php
+++ b/output/Html.php
@@ -146,7 +146,7 @@ abstract class QM_Output_Html extends QM_Output {
 	 * @return string Markup for the column sorter controls.
 	 */
 	protected function build_toggler() {
-		$out = '<button class="qm-toggle" data-on="+" data-off="-" aria-expanded="false">+<span class="screen-reader-text">Toggle button</span></button>';
+		$out = '<button class="qm-toggle" data-on="+" data-off="-" aria-expanded="false">+<span class="screen-reader-text">' . esc_html__( ' Toggle button', 'query-monitor' ) . '</span></button>';
 		return $out;
 	}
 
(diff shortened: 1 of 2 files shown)


rianrietveld:
Hey @johnbillion and @ivdimova 
I think we need `aria-expanded` on the button instead of `aria-pressed`, to tell the screen reader what the `+` means: the content can be expanded and also tell the state. Like we do for hamburger menus. 
Toggle `aria-expanded="false"` and `aria-expanded="true"`

You could also add [aria-controls](https://www.w3.org/WAI/GL/wiki/Using_the_WAI-ARIA_aria-expanded_state_to_mark_expandable_and_collapsible_regions) to link the expandable content to the button.
But the screen reader support for that is questionable. http://www.heydonworks.com/article/aria-controls-is-poop


Also I would suggest extra text on the button (hidden by the 'screen-reader-text' class), to make the button text understandable, not just a `+`

Button list VoiceOver:
Button list VoiceOver

ivdimova:
@rianrietveld @johnbillion this has been changed to aria-expanded, please check it this is ok.

output/Html.php
```diff
@@ -146,7 +146,7 @@ protected function build_sorter() {
 	 * @return string Markup for the column sorter controls.
 	 */
 	protected function build_toggler() {
-		$out = '<button class="qm-toggle" data-on="+" data-off="-" aria-pressed="false">+</button>';
+		$out = '<button class="qm-toggle" data-on="+" data-off="-" aria-expanded="false">+<span class="screen-reader-text">Toggle button</span></button>';
```
rianrietveld: Hey @ivdimova, add a space before the Toggle like `<span class="screen-reader-text"> Toggle button</span>`
That way it's pronounces better.
Then It's ok for me.

The text still isn't very unique, but it's better than it was.

johnbillion: Thanks Rian. @ivdimova Can you also add the missing i18n that we just chatted about?

ivdimova:
Ok to merge now @johnbillion?

Issue #291, opened 2017-12-14
Switch to client side templating for HTML output


johnbillion:
The server-side generation of QM's HTML output leaves a lot to be desired. If the user doesn't interact with QM then all that rendering happens for nothing. When the output gets very large, the delay in generating the HTML becomes noticeable.

QM should switch to client side template rendering ~~via Underscore templates~~. The rendering only needs to occur at the moment when the user clicks an item in QM's admin toolbar menu and the data can be populated with a JSON representation of the collected data.

~~I've decided on Underscore templates because it's bundled with WordPress and means it's the least likely to have problems with conflicts with other libraries.~~

This leads to several benefits, for example reusable templates that can be used outside of QM's included output mechanism. One example could be rendering of QM sessions stored elsewhere, playback of recent sessions, full or partial output for Ajax or REST API responses, server-side rendering of ~~Underscore~~ templates using Node, etc.

Having QM's data available as JSON on the page also opens up some interesting use cases. The data can be accessed via the browser's developer console, data could be sent off client side for logging somewhere else or used as part of analytics, and a headless browser could automatically collect performance stats as part of an automated performance budget monitoring tool.

~~First PR: #292~~  
Current PR: #483

johnbillion:
Latest PR: #483

lkraav:
Heya @johnbillion hopefully all is well.

I'm looking to get something going for scripts and styles handle count totals output in a JSON structure for continous monitoring.

It seems #483 still has ways to go.

Is there a way to reach the above goal with current QM architecture, how would you recommend approaching it?

johnbillion:
Closing in favour of #1068.

PR #292, opened 2017-12-14
Implement client side templates

johnbillion:
See #291.

This begins the process of switching to client side Underscore templates while retaining the ability for any outputter to continue using server side HTML generation.

The first panel to be switched over to an Underscore template is the `Admin Screen` panel.

### DIFF
--- a/dispatchers/Html.php
+++ b/dispatchers/Html.php
@@ -204,7 +204,16 @@ class QM_Dispatcher_Html extends QM_Dispatcher {
 			$timer = new QM_Timer;
 			$timer->start();
 
-			$output->output();
+			if ( method_exists( $output, 'template' ) ) {
+				?>
+				<script type="text/html" id="tmpl-qm-<?php echo esc_attr( $id ); ?>">
+					<?php $output->template(); ?>
+				</script>
+				<div id="qm-out-<?php echo esc_attr( $id ); ?>"></div>
+				<?php
+			} else {
+				$output->output();
+			}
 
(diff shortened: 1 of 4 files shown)


dispatchers/Html.php
```diff
@@ -154,7 +154,10 @@ public function enqueue_assets() {
 		wp_enqueue_script(
 			'query-monitor',
 			$this->qm->plugin_url( 'assets/query-monitor.js' ),
-			array( 'jquery' ),
+			array(
+				'jquery',
+				'wp-util',
```
shadyvb: I think you should either add `underscore` or remove `jquery`, as `wp-util` depends on both.

johnbillion: I think I'll add Underscore so the dependencies are more explicit.

tomjn:
Doing this via React or Vue might make more sense, QM can output the entire debug info to be displayed as JSON and it be passed down the hierarchy, the only things that need state being hiding one panel and showing another.

Also, using underscores templates brings with it the problem that content security policies can't remove unsafe eval etc on the frontend

johnbillion:
I wanted to avoid loading another dependency that isn't included in WordPress, but you're right that React or Vue would be a better choice.

This is on hold at the moment anyway.

lkraav:
> I wanted to avoid loading another dependency that isn't included in WordPress, but you're right that React or Vue would be a better choice.

React will be shipping w/ Gutenberg soon enough. At least the dependency shipping problem looks solved.

Issue #282, opened 2017-11-12
PHP functions running times?


eddr:
Hi man, great plugin. It's really next level of WP development

Wanted to ask - any chance you'll add in the future performance indicators for the PHP functions?

EDIT: I see now that you said you'll probably add such functionality

johnbillion:
Thanks for the suggestion.

This isn't something that I'm planning on adding to QM, although it is possible to do via ticks in PHP but the overhead is significant (which skews the profiling). It would need to be an opt-in feature, or an add-on plugin.

What I _am_ planning on adding is an API so developers can perform profiling of their own code using something along the lines of this:

```php
do_action( 'qm/start', 'foo' );
my_long_running_function();
do_action( 'qm/stop', 'foo' );
```

It would support laps too:

```php
do_action( 'qm/start', 'foo' );
foreach ( $items as $item ) {
	do_something();
	do_action( 'qm/lap', 'foo' );
}
do_action( 'qm/stop', 'foo' );
```

The timing and memory usage information for this profile would then be output in a panel in QM.

No ETA for this feature just yet.

crstauf:
@eddr [QMX](https://github.com/crstauf/query-monitor-extend) supports adding benchmarks in your code, and displaying them in QM (though I **strongly** look forward to @johnbillion's implementing into QM).

eddr:
@crstauf thanks man, I appreciate the help!
I would probably wait

johnbillion:
First pass of this is now in the develop branch. See #301.

PR #301, opened 2018-01-12
Adding option to track php functions time with adding actions in the …

ivdimova:
…code that will be profilled
For #282

### DIFF
--- a/collectors/timing.php
+++ b/collectors/timing.php
@@ -1,0 +1,63 @@
+<?php
+/*
+Copyright 2009-2017 John Blackbourn
+
+This program is free software; you can redistribute it and/or modify
+it under the terms of the GNU General Public License as published by
+the Free Software Foundation; either version 2 of the License, or
+(at your option) any later version.
+
+This program is distributed in the hope that it will be useful,
+but WITHOUT ANY WARRANTY; without even the implied warranty of
+MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+GNU General Public License for more details.
+
+*/
+
+class QM_Collector_Timing extends QM_Collector {
+
+	public $id = 'timing';
+	private $start = null;
+	private $stop = null;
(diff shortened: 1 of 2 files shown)


johnbillion:
Yep this is exactly the sort of thing I was thinking, thanks. A few points:

* The `QM_Timer` class can be used in place of manually calling `microtime()`. It supports starting, lapping, and stopping a timer and includes the time and memory usage. You can see this in use in the self-awareness output in the `develop` branch.
* Probably will need to handle timers that were started but not stopped, with a useful warning message.
* We'll need to consider overlapping timers. The `start` and `stop` properties in the collector might need to be converted to associative arrays keyed by the function name. Example that needs to be handled:

```php
do_action( 'qm/start', 'function_1' );
do_action( 'qm/start', 'function_2' );
...
do_action( 'qm/stop', 'function_1' );
do_action( 'qm/stop', 'function_2' );
```

eddr:
Will you have to manually add "do_action" for each function you want to measure? if so, that's fine, but what about functions inside classes or if you want to get a measurement of the whole system at once?

crstauf:
@eddr Interesting idea: a single location where you could add all of your timers, rather than having to manually add into each function/method. 

@johnbillion Possible to add `qm/start`/`qm/stop` onto a hook, and if there's no param (function name) provided, assume `current_action()`?

johnbillion:
For now you'll need to manually call the actions (don't worry, I'm writing up some documentation for this feature), but some sort of automated measurements is a good idea and I'll look at it at a later date.

@crstauf Good idea, I like it. #309.

crstauf:
@johnbillion has the documentation for these functions been created? I'm attempting to implement, but doesn't look like lapping is being displayed in the output?

crstauf:
Addressed in https://github.com/johnbillion/query-monitor/issues/337#issuecomment-387065076

PR #313, opened 2018-02-24
MySQL 5.7.20 query_cache_* deprecated

EvanHerman:
MySQL 5.7.20 deprecated `query_cache_*` which results in false warnings.

See: https://dev.mysql.com/doc/refman/5.7/en/query-cache-configuration.html

This PR unsets the values in the `$mysql_vars` array when on a site running MySQL 5.7.20 or later.


johnbillion:
Thanks for the PR!

PR #334, opened 2018-04-26
Modify check for is_main_query

### LINKED ISSUE #333: PHP "Undefined index" notices firing for user queries
When viewing something like `wp-admin/users.php`, the plugin emits notices like:

```
Undefined index: is_main_query
```

This seems to be because the checks for `$row['is_main_query']` in `output/html/db_queries.php` assume that the index at least exists, which it doesn't in the case of a user query.

jacklenox:
Fixes #333.

### DIFF
--- a/output/html/db_queries.php
+++ b/output/html/db_queries.php
@@ -351,7 +351,8 @@ class QM_Output_Html_DB_Queries extends QM_Output_Html {
 		if ( isset( $cols['caller'] ) ) {
 			$row_attr['data-qm-caller'] = $row['caller_name'];
 
-			if ( $row['is_main_query'] ) {
+			$is_main_query = isset( $row['is_main_query'] ) ? $row['is_main_query'] : false;
+			if ( $is_main_query ) {
 				$row_attr['data-qm-caller'] .= ' qm-main-query';
 			}
 		}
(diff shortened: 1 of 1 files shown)


gmazzap:
Any news on this?

My logs are exploding with `"Undefined index: is_main_query"`...

justnorris:
@johnbillion It'd be really great to get this merged in soon as it's causing a lot of unnecessary notices in the UI when PHP errors are displayed (like in the typical dev-environment)

johnbillion:
I'm gonna get this fixed in 3.1 but probably by fixing the root cause instead.

@justnorris Can you confirm that you're using a plugin such as HyperDB, or another `wp-content/db.php` database dropin?

justnorris:
Yeah, this is happening on VIP Go and I think it might have to do with the fact that VIP Go query collector is a bit out of sync with QM. I'm doing a diff on the files right now:

https://github.com/Automattic/vip-go-mu-plugins/blob/7ddbaeed75da5ae425ed79ba99c8e49edf9c7246/vip-helpers/vip-query-monitor.php

https://github.com/johnbillion/query-monitor/blob/master/collectors/db_queries.php

johnbillion:
Ah yes that's most likely the cause, thanks.

johnbillion:
See #333 .

PR #351, opened 2018-07-12
Template Part file string normalization

JasonBusse:
The template parts were not showing up on my machine because the file strings returned by get_included_files() have \ instead of / directory separators which caused the string comparisons to fail. Using wp_normalize_path ensures the separators match.


johnbillion:
Thanks for the PR!

Issue #361, opened 2018-08-21
Panel menu UL is showing disc list styles which breaks layout


chrisblakley:
I noticed this when Query Monitor was first updated to 3 and it has persisted since then after a few plugin updates. I'm not sure if others are seeing it, but I'll make a pull request to explicitly declare `list-style: none;` on those `<ul>` elements.

Before:
![screen shot 2018-08-21 at 8 54 06 am](https://user-images.githubusercontent.com/6349678/44402740-a1fc4080-a520-11e8-8fb9-93a8b752f6b9.png)

After:
![screen shot 2018-08-21 at 8 56 58 am](https://user-images.githubusercontent.com/6349678/44402748-a58fc780-a520-11e8-8514-71b18b441fe4.png)

johnbillion:
Thanks for the report! Out of interest, which theme are you using?

chrisblakley:
I have a custom developed theme that I've used with Query Monitor for years- here are some links for reference if that helps at all:
- https://github.com/chrisblakley/Nebula
- https://gearside.com/nebula

The thought that it could be a theme conflict did cross my mind (it's certainly happened before), but that's why I thought making an explicit declaration within QM would be the way to go.

johnbillion:
Yeah adding the reset to QM is definitely needed. I've been testing QM against a bunch of the most popular themes but styles like this will keep popping up :-).

johnbillion:
Fixed in #362.

PR #362, opened 2018-08-21
Explicitly remove disc list style on panel menus

chrisblakley:
Not sure if this is the right file (didn't see a Sass file in this panel UI branch), but this is the simple fix for what I've been seeing described in this issue: https://github.com/johnbillion/query-monitor/issues/361


johnbillion:
Thanks!

PR #364, opened 2018-09-05
Build/PHPCS: use PHPCompatibilityWP

jrfnl:
This PR:
* Switches the repo over to use `PHPCompatibilityWP` rather than `PHPCompatibility`.
    As this repo and ruleset is a WordPress projects, using the `PHPCompatibilityWP` ruleset is the better choice to prevent false positives for PHP functions and constants which are back-filled by WP.
* Actually adds the `PHPCompatibilityWP` standard to the custom ruleset to activate it.

### References:
* https://github.com/PHPCompatibility/PHPCompatibilityWP
* https://github.com/PHPCompatibility/PHPCompatibility/issues/688
* https://github.com/PHPCompatibility/PHPCompatibility/releases/tag/8.2.0

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -21,7 +21,7 @@
 	},
 	"require-dev" : {
 		"squizlabs/php_codesniffer": "^3.2",
-		"phpcompatibility/php-compatibility": "~8.2",
+		"phpcompatibility/phpcompatibility-wp": "^1.0",
 		"wp-coding-standards/wpcs": "0.13",
 		"phpunit/phpunit": "^5"
   }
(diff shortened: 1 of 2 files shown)


jrfnl:
@johnbillion This PR is outdated by now as PHPCompatibilityWP 2.0 has come out since.
I would also recommend removing the `installed_paths` config from the ruleset and using the `dealerdirect/phpcodesniffer-composer-installer` composer plugin instead, especially as `PHPCompatibilityWP 2.0` will have more dependencies on other sub-standards.

I'd be happy to make the necessary changes, but it would be good if the PR could then be merged reasonably quickly to prevent it going out of date and into merge conflict again.

johnbillion:
Yep thanks. I've just done the update in ca12075.

jrfnl:
Great. LGTM :+1:

PR #367, opened 2018-10-10
Bugfix: min-width CSS for buttons

Daronspence:
I had a client project that used `min-width` for the global button styles and I noticed it was affecting the QM styles as well.

I'm not sure if the output is correct, but hopefully this will point you in the right direction if I messed up. :)


johnbillion:
Thanks!

PR #388, opened 2019-01-22
Dropped OPCache from Object Cache backends

### LINKED ISSUE #372: Incorrectly states OPCache is not in use
As the title says, Query Monitor incorrectly reports that the opcache extension is not in use. It looks like the code just lists the installed caching extensions if an object cache is not in use.

opcache-in-use

https://github.com/johnbillion/query-monitor/blob/dae2820587dea95a169b502cc85b49e5a9195943/output/html/overview.php#L182-L187

Rarst:
Fixes #372

### DIFF
--- a/collectors/cache.php
+++ b/collectors/cache.php
@@ -79,7 +79,6 @@ class QM_Collector_Cache extends QM_Collector {
 				'Memcache'     => 'Memcache',
 				'Memcached'    => 'Memcached',
 				'Redis'        => 'Redis',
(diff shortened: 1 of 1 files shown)


johnbillion:
Thanks for the PR, but I've already fixed this in the `develop` branch by splitting out the opcode cache handling from the object cache handling. My mistake for not referencing the issue anywhere.

https://github.com/johnbillion/query-monitor/commit/9cfac7499e77c8602bef89d6c9bbc350858590c4

PR #392, opened 2019-02-06
Show panel on side

timiwahalahti:
Sometimes when working on a wide monitor, I dream of having a QM panel in the side rather than on the bottom of the site.

This is a simple PR to demonstrate how it would look. If the idea is liked and John sees it something that QM should contain, I'll work on this to make better PR.

### TODO
- disable height scaling, add width scaling
- default adminbar height for `top` positioning, but ensure with JS that it is correct
- move styles to scss for build process
- add toggle side/bottom icon next to the close icon

image


johnbillion:
NICE! I definitely like the idea of this. Thanks for working on it.

I've worked quite hard to get QM to scroll horizontally without squashing things up too much, hopefully this helps with the usability of a narrow QM panel on the side. If you see any issues with very narrow columns, do let me know.

P.S. my long germ goal is to produce a browser extension for QM that handles the UI, and then QM would only output a JSON blob to the page that the extension interprets. But this is a long way off, and the standard HTML view within the page would still remain.

johnbillion:
Don't worry about the branch conflicts for now, I'm going to be making some more Sass and CSS tweaks in the develop branch. We can fix the conflicts when this gets nearer to being ready.

timiwahalahti:
Basically works quite nice now, there are few things to fix:
- hide position toggle icon when on a narrow screen
- on window load, check if the screen is narrow
- do not reset width when the position is changed back to the side
- consider if there is better dashicon than `image-rotate-left`

timiwahalahti:
Added checks to handle situations where the window is narrow or width is smaller than stored panel width. Fixes the width reset issue on the position change.

Personally, I don't like the `image-rotate-left` dashicon indicating the position switching but can't figure out better either. Maybe toggle between two: when panel is in default position show `dashicons-image-flip-horizontal` and when on side, show `dashicons-image-flip-vertical`. Icon toggling might make the UI a bit confusing tho.

johnbillion:
Thanks for working on this @timiwahalahti . I'm going to make some further tweaks.

I've just pushed some commits which means when QM is shown on the side, it reuses the same layout that gets used on narrow viewports. The main menu is hidden, and replaced with a dropdown menu at the top of the panel. This increases the available horizontal space.

What do you think?

johnbillion:
I've added a resize grabber on the side of the panel too, so its resize controls are more intuitive.

I'm happy with this and I'm going to merge it in. Many thanks for this neat feature!

timiwahalahti:
Awesome! Thanks for tweaking this to work even better and merging :)

Issue #394, opened 2019-02-07
Authentication cookie does not work in Pantheon hosting


carl-alberto:
It has been a mystery for me and a lot of Pantheon users that the authentication cookie is not working as it is very essential in troubleshooting logged out users.

This issue seemed to be similar to another plugin that we were able to find out the culprit https://wordpress.org/support/topic/troubleshooting-mode-in-pantheon-hosting-do-not-work/

It seems the cookie is being stripped by the Pantheon’s Varnish cache layer not letting it pass as per naming specifications defined in the platform’s Varnish configuration: https://pantheon.io/docs/cookies/#cache-busting-cookies, any cookie that do not have a `wp-` or `wordpress_` prefix is being stripped.

I tried it in my test environment and I was able to make the authentication cookie work by replacing the code that sets the cookie from `query_monitor_` to `wp-query_monitor_` or `wordpress_query_monitor_`.

Tried to open an issue here as this seems to be code related and might be a good place to start with in relation to: https://wordpress.org/support/topic/authentication-cookie-does-not-work-in-pantheon-hosting/#post-11180920

crstauf:
@carl-alberto As far as you know, Pantheon does not allow cache exclusions by cookie then, is that right?

crstauf:
The concern I have here (and it may be completely unfounded) is that this is technically a limitation of Pantheon: several plugins use cookies that do not have the `wp-` prefix, but seemingly, Pantheon does not permit cache exclusions by custom cookies. This seemingly then means plugins must adhere to Pantheon's established rules, which just seems crazy to expect all developers everywhere to know about Pantheon: "oh right, we have to prefix our cookies with `wp-` because of Pantheon".

Again, I may be way off here, but I've seen this combo before, and the usual conclusion is for the host to become more accommodating.

carl-alberto:
@crstauf As of the moment, Pantheon doesn't have a facility to exclude cookies only by urls so it might be a good feature request to propose to accommodate more plugins/users in the platform. You have a valid point and agree to it. Just trying to explore the possibility of the plugin author to have a prefix, but if not feasible, will move on to other possible solutions.

johnbillion:
I think it's pretty reasonable to change QM's cookie to reduce friction on platforms such as Pantheon that have this sort of Varnish configuration. I know that WordPress.com VIP also has a similar requirement, IIRC.

carl-alberto:
[...]

PR #395, opened 2019-02-07
Proposed cookie prefix (wp-) to have hosting compatibility in Pantheon

carl-alberto:
Fixes: https://github.com/johnbillion/query-monitor/issues/394


johnbillion:
Thanks!

PR #407, opened 2019-02-16
Ability to time individual hooks

### LINKED ISSUE #309: Ability to time individual hooks
Via @crstauf on https://github.com/johnbillion/query-monitor/pull/301#issuecomment-364177323:

Possible to add `qm/start`/`qm/stop` onto a hook, and if there's no param (function name) provided, assume `current_action()`?

QM could enable this via a wrapper function such as `qm_start()` and `qm_stop()` which would then be used like this:

```php
add_action( 'after_setup_theme', 'qm_start', 0 );
add_action( 'after_setup_theme', 'qm_stop', 999 );
```

Maybe even only require the start action and QM adds the stop action itself.

crstauf:
Closes #309.


crstauf:
@johnbillion I know QM is in feature freeze right now, but given that this is from 2019, is there anything that needs to be done that kept this from being accepted back then?

Issue #409, opened 2019-02-18
Distinguish between related hooks that fired and those that didn't


johnbillion:
The `Hooks in Use` submenu for each panel can list a large number of hooks, many of which won't have fired during the current request.

Need a way to distinguish between those that fired and those that didn't

crstauf:
@johnbillion How can/are you replicating this?

johnbillion:
I got confused here with some exploratory work that I was doing locally. This panel definitely only shows hooks that have fired during the current request. Mea culpa.

PR #412, opened 2019-02-21
Concerned filters for the Queries panel

johnbillion:
Probably too much in here. Need to decide how to make this useful rather than potentially overwhelming.


crstauf:
@johnbillion Do you think #409 would reduce the number of hooks enough to not be overwhelming?

PR #417, opened 2019-02-27
updated CSS to avoid conflicts with themes using ul, nav, and li class names

emaildano:
## Issue
I recently ran into an issue with the CSS of a WordPress theme affecting layout of the Query Monitor panel. Specifically, class names in the Theme using `ul`, `li`, and `nav`.

## Proposed Solution
Update the following CSS classes to maintain the layout of the Query Monitor panel.

```
#query-monitor-main #qm-panel-menu ul {
  width: auto !important;
  height: auto !important;
}

#query-monitor-main #qm-panel-menu li {
  height: auto !important;
}

#query-monitor-main #qm-panel-menu {
  height: auto !important;
}
```

My PR details this, but it also included some formatting fixes which are based purely on VS Code. Those are not essential to this PR or fix.

### DIFF
--- a/assets/_styles.scss
+++ b/assets/_styles.scss
@@ -355,6 +355,7 @@ body.admin-color-light #wp-admin-bar-query-monitor:not(.qm-all-clear):not(:hover
 		flex-shrink: 0 !important;
 		overflow-y:  scroll !important;
 		overscroll-behavior: contain !important;
+		height: auto !important;
 	}
 
 	#qm-panel-menu ul {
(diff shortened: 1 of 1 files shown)


emaildano:
screen shot 2019-02-27 at 1 39 04 pm

For reference, this is the issue I was seeing. Adding that CSS was able to correct the layout.

lkraav:
Real solution here is Web Components ShadowDOM implementation.

emaildano:
@lkraav Agreed. That's the solution, but maybe we just need a fix for now.

@johnbillion Is that something on your radar? I'd def be interested in creating a PR to update Query Monitor to use ShadowDOM if you're interested.

Unless @lkraav is interested in writing a PR?

lkraav:
Would love to but no extra time right now. 

Main thing is to choose a light framework to build with, perhaps LitElement, or just write the boilerplate JS ourselves (if target is a only single container component).

johnbillion:
Thanks for the PR.

I'm not interested in a shadow DOM solution at the moment, I've got lots more stuff that's higher priority.

emaildano:
@johnbillion Thanks for merging it! I'm happy I was able to contribute.

PR #434, opened 2019-03-22
Define `$return['file']` and `$return['line']`

### LINKED ISSUE #433: 3.3.3: WP_Die $filtered_trace PHP Notice missing index 'file' and 'line'
I am seeing PHP Notices for these missing indexes when using the Debug Quick Look plugin to view the debug.log.

afragen:
Make sure `$return['file']` and `$return['line']` are defined. 

Fixes #433

### DIFF
--- a/classes/Backtrace.php
+++ b/classes/Backtrace.php
@@ -326,6 +326,8 @@ class QM_Backtrace {
 
 			$return['calling_file'] = $this->calling_file;
 			$return['calling_line'] = $this->calling_line;
+			$return['file']         = isset( $trace['file'] ) ? $trace['file'] : null;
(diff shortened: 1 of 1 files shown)


johnbillion:
Thanks for the PR @afragen . This is probably papering over a lower level issue, so I'd like to see if I can find the root cause of the missing properties.

afragen:
The error shows up in WP_Die. I traced it back and it seemed as if the `filter_trace()` was a good place to fix. I will certainly defer to your preference. 

Thanks for fixing this and thanks for 10 years of QM. ❤️

johnbillion:
Thanks for the PR but I'll close this off in favour of #602 which is a more fundamental fix.

PR #437, opened 2019-04-01
Plugin compatibility check may result in empty $filtered_trace

afragen:
In WP 5.2 a plugin that doesn't meet the minimum PHP requirements will throw a WP_Die error. When tracing this in QM there there is only a single item in $filtered_trace. After the `array_pop( $filtered_trace );` the value of $filtered_trace is an empty array.

This PR accounts for that edge case.

### DIFF
--- a/dispatchers/WP_Die.php
+++ b/dispatchers/WP_Die.php
@@ -46,6 +46,10 @@ class QM_Dispatcher_WP_Die extends QM_Dispatcher {
 			$stack[] = QM_Output_Html::output_filename( $item['display'], $item['file'], $item['line'] );
 		}
 
+		if ( empty( $filtered_trace ) ) {
+			return;
+		}
+
 		if ( isset( $filtered_trace[ $i - 1 ] ) ) {
 			$culprit = $filtered_trace[ $i - 1 ];
 		} else {


johnbillion:
Thanks for the PR, I'll take a look!

johnbillion:
@afragen Can you let me know how to reproduce this `wp_die()` error please? Cheers!

afragen:
I have my WP-Debugging plugin active which uses a modified version of Norcross’s Debug Quick Look plugin to view the debug log. 

I’m not certain how to replicate this all the time but I’ll see if I can create something to do this.

afragen:
I found out where I see this. If the patch for https://core.trac.wordpress.org/ticket/46938 is accepted, trying to activate a plugin that has the appropriate plugin headers set and is incompatible will throw this error.

johnbillion:
This doesn't seem to be a problem now that this has been merged into core, or at least I can't reproduce it on that `wp_die()` screen with an incompatible plugin.

I'll close this off, but let me know if it's still reproduceable somehow.

afragen:
I haven’t seen this again. So I think it’s safe as closed.

PR #444, opened 2019-05-22
Add CLI command to create the symlink to the db file.

msaggiorato:
Dragging inspiration from other plugins that need a symlink created (like wp-redis), i'm proposing to have a similar command for easily enabling the extended features of Query Monitor.

I've created the class with the same structure as other classes loaded early.

Command added is `wp qm enable`

### DIFF
--- a/classes/CLI.php
+++ b/classes/CLI.php
@@ -1,0 +1,94 @@
+<?php
+/**
+ * Plugin CLI command.
+ *
+ * @package query-monitor
+ */
+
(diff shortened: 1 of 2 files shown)


johnbillion:
Great idea, thanks! I won't have a ton of time to review this for a while, but I'll definitely get to it as soon as I can.

msaggiorato:
Thanks for the feedback @johnbillion! Wouldn't having a hard link make these installations not work? Or does QM auto corrects this on the fly?

PR #452, opened 2019-06-07
Add headers and other request data to response

rmccue:
Adds a "raw request data" collector, then uses this in various places across the output, plus adds some new headers panels.

## Overview panel
Screen Shot 2019-06-07 at 15 14 18

## Request panel
Screen Shot 2019-06-07 at 15 10 10

## Request Headers panel (new!)
Screen Shot 2019-06-07 at 15 10 19

## Response Headers panel (new)
Screen Shot 2019-06-07 at 15 10 25


collectors/raw_request.php
```diff
@@ -0,0 +1,84 @@
+<?php
+
+class QM_Collector_Raw_Request extends QM_Collector {
+
+	public $id = 'raw_request';
```
rmccue: I was unsure on exactly what to call this collector; "request" would make the most sense, but it's an overloaded term already being used.

collectors/raw_request.php
```diff
@@ -0,0 +1,84 @@
...
+
+	/**
+	 * Extracts headers from a PHP-style $_SERVER array.
+	 *
+	 * From WP_REST_Server::get_headers()
+	 *
+	 * @param array $server Associative array similar to `$_SERVER`.
+	 * @return array Headers extracted from the input.
+	 */
+	protected function get_headers( $server ) {
```
rmccue: I copied this across, but `rest_get_server()->get_headers()` is equivalent; it forces the API to load though, so not the greatest.

johnbillion:
As an FYI the reason this sort of request and response information isn't exposed by QM currently is that it's all available in the browser developer tools.

Given that I'm looking to add support for persistence to QM I'm happy to add this in.

rmccue:
Worth noting it’s also pretty useful for detecting whether proxies/reverse proxies are altering headers.

PR #454, opened 2019-06-10
Add ability to replace Collectors instance

### LINKED ISSUE #447: Add ability to replace singleton instances
I'm working on the ability to persist request data for QM, which is pretty easy thanks to the separation of collectors from dispatchers/outputters. The only thing that makes this a little bit of a pain is the use of the `QM_Collectors` singleton.

This singleton instance is stored as a static variable within `QM_Collectors::init`, and cannot be replaced. This means I can't replace the collected data. Currently, I'm relying on the load order of my plugin before QM to declare the class before QM does, but this is brittle.

Two possible solutions:
1. Add an optional parameter to `::init()` with an instance of the class, and replace the static variable with that instance if provided.
2. Store the instance as a static class variable and add a method to replace it.

rmccue:
Fixes #447.

Two notes:

1. `Dispatchers` follows the same pattern; I haven't patched it here, but can if you'd like.
2. I'm using `static::` here, since WP's minimum version now supports it. The rest of the code is using `self::` however; I can update this to be consistent in either direction if you'd like.

### DIFF
--- a/classes/Collectors.php
+++ b/classes/Collectors.php
@@ -7,6 +7,7 @@
 
 if ( ! class_exists( 'QM_Collectors' ) ) {
 class QM_Collectors implements IteratorAggregate {
+	private static $instance = null;
 
 	private $items     = array();
 	private $processed = false;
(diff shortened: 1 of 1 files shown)


rmccue:
@johnbillion Any thoughts here on whether this is the approach you want to take?

johnbillion:
I'm going to close this off and revisit it once #483 is in place. Cheers.

PR #465, opened 2019-08-30
Filter QM_Util::get_file_dirs to add 3rd-party components

### LINKED ISSUE #450: Add ability to filter component
It'd be great if there was a filter in `QM_Util::get_file_component()` (and `QM_Util::get_file_dirs()`). This could be used by third-parties to group components more easily and accurately with custom setups.

For example, the vip-plugin support could be handled entirely via filters with this approach.

abuyoyo:
fixes #450 

- Added the filters. They work as expected with a custom plugin directory I have installed. Doesn't seem to break anything.
- Did not remove any of the hard-coded external directories already in QM_Util. Frankly I've no idea where they would go in the code.
- Filtered directories are added first in `get_file_dirs` because of the foreach-break in `get_file_component` (we can't know where the 3rd-party directory is installed - might be a subdirectory of a normal plugin in `plugins`. Very likely it's a subdirectory of ABSPATH or WP_CONTENT_DIR etc.)

### DIFF
--- a/classes/Util.php
+++ b/classes/Util.php
@@ -69,23 +69,31 @@ class QM_Util {
 
 	public static function get_file_dirs() {
 		if ( empty( self::$file_dirs ) ) {
-			self::$file_dirs['plugin']     = self::standard_dir( WP_PLUGIN_DIR );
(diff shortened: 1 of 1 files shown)


johnbillion:
Thanks @abuyoyo

Issue #474, opened 2019-10-01
Feature request: add filter for http request status codes to flag as errors


pbiron:
There is currently a filter to control the PHP error level for specific plugins/themes (see [Silencing Errors from Certain Plugins and Themes in Query Monitor](https://querymonitor.com/blog/2018/07/silencing-errors-from-plugins-and-themes/)) that stops the "coloured notification in the admin toolbar" for when those plugins/themes generate certain PHP errors.

It would be nice if there were a similar filter for HTTP response status codes produced by calls to `wp_remote_xxx()`.

I'm writing a plugin that does several `http_remote_head()` calls, most of which will return 404s...and that is perfectly OK and expected for the operation of that plugin.  I want Query Monitor to log the requests, but not flag these specific 404s as "errors" (i.e., not turn the QM admin bar item red).

If there were such a filter, I could bracket these `wp_remote_head()` calls with `{add,remove}_filter()`  calls that hook/unhook that filter.

johnbillion:
I can probably add something in for this.

In the meantime, there's a `qm/collect/silent_http_errors` filter which is used to silence errors returned by the Airplane Mode plugin and errors returned when the request is blocked. If your request returns a specific error code you could use it in this filter to silence the error.

pbiron:
Thanx!

Yeah, I saw that filter and at first thought it was promising.  But then I realized it was for `WP_Error` codes.

I guess until a version of QM is released that has a filter like I'm asking for I could wrap my calls to `wp_remote_head()` in a method and have that method return a specific `WP_Error` if the `HEAD` results in a 404 and then use `qm/collect/silent_http_errors` to silence that call.

What I'm thinking for the new filter is something like:

```
$silent_error_statuses = apply_filters( 'qm/collect/silent_http_error_statuses', array());
...
if ( $http['type'] >= 400 && ( empty( $silent_error_statuses ) || ! in_array( $http['type'], $silent_error_statuses ) ) ) {
	$this->data['errors']['warning'][] = $key;
}

I can do a PR if you'd like.
```

pbiron:
> I guess until a version of QM is released that has a filter like I'm asking for I could wrap my calls to `wp_remote_head()` in a method and have that method return a specific `WP_Error` if the `HEAD` results in a 404 and then use `qm/collect/silent_http_errors` to silence that call.

now that I think about it, that workaround won't work :-(

johnbillion:
[...]

PR #475, opened 2019-10-02
Add new qm/collect/silent_http_error_statuses filter to silence specific HTTP API requests that result in 400-500 HTTP status codes

pbiron:
Fixes: #474

The new filter is similar to the existing `qm/collect/silent_http_errors`.

The difference are:

1. it is called separately for each `$http` object collected.
2. it is passed the `$http` object so that plugins can decide on a request-by-request basis which HTTP status codes they want to silence.

A sample use of this new filter might be something like:

```
add_filter(
	'qm/collect/silent_http_error_statuses',
	array( $this, 'my_plugin_qm_silence_404s' ), 10, 2
);

function my_plugin_qm_silence_404s( $http_statuses, $http ) {
	if ( in_array( $http['url'], array( 'https://some-host.com/foo', 'https://another-host.com/bar' ) ) &&
			 wp_list_filter( $http['trace']->get_trace(), array( 'function' => 'my_plugin_ping_sites' ) ) ) {
		// tell QM that a 404 is OK in this case.
		$http_statuses[] = 404;
	}

	return $http_statuses;
}

function my_plugin_ping_sites() {
	// these requests should be silenced.
	wp_remote_head( 'https://some-host.com/foo' );
	wp_remote_head( 'https://another-host.com/bar' );
	// this request should not be silenced.
	wp_remote_head( 'https://yet-another-host.com/baz' );
}

my_plugin_ping_sites();

function my_plugin_ping_sites_again() {
	// these requests should not be silenced, even tho they are to the same URLs
	// that are silenced in my_plugin_ping_sites().
	wp_remote_head( 'https://somehost.com/foo' );
	wp_remote_head( 'https://anotherhost.com/bar' );
	// this request should not be silenced.
	wp_remote_head( 'https://yet-another-host.com/baz' );
}

my_plugin_ping_sites_again();

```

Also, on the DocBlock for the new filter, the `@since` tag is `vnext`, because I'm not sure what version of QM this might make it into; and in the hash for the `$http` param, I didn't do a sub-hashes for the array args (e.g., `$response` and `$info`).  I can do that if you'd like.

### DIFF
--- a/collectors/http.php
+++ b/collectors/http.php
@@ -255,7 +255,38 @@ class QM_Collector_HTTP extends QM_Collector {
 			} else {
 				$http['type'] = intval( wp_remote_retrieve_response_code( $http['response'] ) );
 				if ( $http['type'] >= 400 ) {
-					$this->data['errors']['warning'][] = $key;
+					/**
+					 * List of HTTP error status codes to ignore.
+					 *
+					 * @since vnext
+					 *
+					 * @param array $http_statuses Array of HTTP error statuses to silence.
+					 *              Default: empty array.
+					 * @param array $http {
+					 *     The QM "HTTP" request object.
+					 *
+					 *     @type string $url The URL for the HTTP request.
+					 *     @type array  $args The `$args` for the HTTP request.
+					 *     @type float  $start The start time for the request.
+					 *     @type class  $trace The QM_Backtrace object for the request.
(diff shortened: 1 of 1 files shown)


pbiron:
Closing this PR and will submit another with an alternate approach shortly.

PR #479, opened 2019-10-16
Feature/editor selector

tomjn:
First go at adding a built in editor selector to the config menu, when chosen it sets a cookie that then adds a pre-registered editor protocol handler. Probably needs some extra polish, but I'd like a set of eyes to make sure this is the preferred approach

### DIFF
--- a/dispatchers/Html.php
+++ b/dispatchers/Html.php
@@ -416,17 +416,20 @@ class QM_Dispatcher_Html extends QM_Dispatcher {
 
 		echo '<p>';
 		echo '<select id="qm-editor-select" name="qm-editor-select" class="qm-filter">';
+
 		$editors = array(
-			'Default/XDebug'     => '',
-			'PHPStorm'           => 'phpstorm',
+			'Default/Xdebug'     => '',
 			'Atom'               => 'atom',
 			'Netbeans'           => 'netbeans',
-			'Visual Studio Code' => 'vscode',
+			'PhpStorm'           => 'phpstorm',
 			'Sublime Text'       => 'sublime',
+			'Visual Studio Code' => 'vscode',
 		);
+
 		foreach ( $editors as $name => $value ) {
 			echo '<option value="' . esc_attr( $value ) . '" ' . selected( $value, $editor, false ) . '>' . esc_html( $name ) . '</option>';
 		}
+
 		echo '</select>';
 		echo '</p><p>';
 		echo '<button class="qm-editor-button qm-button">' . esc_html__( 'Set editor cookie', 'query-monitor' ) . '</button>';
--- a/dispatchers/Html.php
+++ b/dispatchers/Html.php
@@ -91,10 +91,11 @@ class QM_Dispatcher_Html extends QM_Dispatcher {
 
 		$expiration = time() + ( 2 * YEAR_IN_SECONDS );
 		$secure     = self::secure_cookie();
+		$editor     = wp_unslash( $_POST['editor'] );
 
-		setcookie( QM_EDITOR_COOKIE, $_POST['editor'], $expiration, COOKIEPATH, COOKIE_DOMAIN,$secure, false  );
+		setcookie( QM_EDITOR_COOKIE, $editor, $expiration, COOKIEPATH, COOKIE_DOMAIN, $secure, false );
 
-		wp_send_json_success( $_POST['editor'] );
+		wp_send_json_success( $editor );
 
 	}
 
--- a/dispatchers/Html.php
+++ b/dispatchers/Html.php
@@ -85,7 +85,7 @@ class QM_Dispatcher_Html extends QM_Dispatcher {
 
 	public function ajax_editor_set() {
 
-		if ( ! self::user_verified() || ! check_ajax_referer( 'qm-editor-set', 'nonce', false ) ) {
+		if ( ! current_user_can( 'view_query_monitor' ) || ! check_ajax_referer( 'qm-editor-set', 'nonce', false ) ) {
 			wp_send_json_error();
 		}
 
--- a/dispatchers/Html.php
+++ b/dispatchers/Html.php
@@ -384,6 +384,34 @@ class QM_Dispatcher_Html extends QM_Dispatcher {
 		echo '<p data-qm-state-visibility="on"><span class="dashicons dashicons-yes qm-dashicons-yes"></span> ' . esc_html__( 'Authentication cookie is set', 'query-monitor' ) . '</p>';
 
 		echo '<p><button class="qm-auth qm-button" data-qm-text-on="' . esc_attr( $text['on'] ) . '" data-qm-text-off="' . esc_attr( $text['off'] ) . '">' . esc_html( $text[ $state ] ) . '</button></p>';
+
+		echo '</section>';
+		echo '</div>';
+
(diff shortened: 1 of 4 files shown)


tomjn:
Screenshot 2019-10-15 at 14 08 26

johnbillion:
I like this and I like the approach. I'll try to find some time over the next week or so to review it properly.

crstauf:
Wondering about the decision to use a cookie; could it also be stored in user's meta for some extended persistence?

I frequently clear cookies, so a more persistent storage would be great. I dont think it should replace the cookie, but in addition to.

Thoughts?

tomjn:
> Wondering about the decision to use a cookie; could it also be stored in user's meta for some extended persistence?

If we store it in user meta, then stack traces don't respect the setting when the user is logged out with an auth cookie.

Ideally, no cookie is needed and it gets stored in local storage, with the JS doing the substitution rather than the PHP

crstauf:
> when the user is logged out with an authorized cookie

Right, that's why I was thinking a combination of the two.

roborourke:
@crstauf good point in regards to things you might do frequently on a dev environment. I don't know if I'd rely on user meta though (can you use query monitor when logged out 🤔). Along with the approach Tom has here it'd be cool if you could define a constant to set the preference. Typically you'd have a local config that isn't committed which would be the place to add it.

johnbillion:
QM respects the `xdebug.file_link_format` directive in php.ini, and also provides a `qm/output/file_link_format` filter to override it. I don't think another constant is necessary.

I like @crstauf 's suggestion about using user meta when it's available and falling back to the cookie if it exists.

johnbillion:
> can you use query monitor when logged out 🤔

Yep! You can set an authentication cookie from QM's Settings panel that let's you use it when logged out or when logged in as a lower level user.

rmccue:
Picking this work up from @tomjn.

> I like @crstauf 's suggestion about using user meta when it's available and falling back to the cookie if it exists.

Should we set the user meta when you click save? If not, how do we set it?

crstauf:
>Shouod we set the user meta when you click save?

@rmccue I think so; putting it in the same JS that sets the localSession key would be ideal, imo.

rmccue:
> @rmccue I think so; putting it in the same JS that sets the localSession key would be ideal, imo.

Not sure why we need the cookie if so, as you could end up in a weird spot where you set it in one browser which sets the cookie and meta, then set in a different browser to something else, but your cookie setting overrides it in the first browser.

We discussed this a bit internally before starting the PR; cookies are IMO the best solution, specifically because they're local to your browser. If I develop on two machines, I might want different editors on each.

I can update if we think user meta is the better way to go, but I don't think it makes sense to do _both_.

Also: updated the PR to now display a message after saving, and note that the change only applies on reload. This may change with #483, but it's a limitation for now.

@johnbillion If you have a chance, we'd like to ship this in a release next week, so would be great if we could get it sorted and released by then :D Totally understand if that's not possible though.

crstauf:
@rmccue I see your point.

@johnbillion [suggested](https://github.com/johnbillion/query-monitor/issues/479#issuecomment-543124239 ) defaulting to the cookie if user meta doesn’t exist, but maybe default to user meta if cookie doesn’t exist?

Essentially we’re trying to resolve repeating steps: 
- if I use one editor across all my computers, it’s preferable for that setting to carry over to all computers
- if I use different editors across all my computers, it’s preferable to not have to change it every time (ex: multiple devs using one WP account)

Using a cookie and defaulting to user meta solves for both situations.

I see one possible issue with that, but I think it would be rare.

Thoughts?

johnbillion:
Yeah happy to get this into the next release. I don't have a strong opinion whether this uses a cookie, user meta, or a combination. I'm hoping it'll be a relatively short-lived feature given the impending switch to client-side rendering, upon which it should be converted to a localStorage preference.

There's some conflicts in the PR, wanna take a look @rmccue ?

rmccue:
Resolved.

As this already uses cookies, I'm going to leave it as that for now, rather than implementing user meta.

dispatchers/Html.php
```diff
@@ -385,6 +403,35 @@ protected function after_output() {
...
+		$editors = [
+			'Default/XDebug'     => '',
+			'PHPStorm'           => 'phpstorm',
+			'Atom'               => 'atom',
+			'Netbeans'           => 'netbeans',
+			'Visual Studio Code' => 'vscode',
+			'Sublime Text'       => 'sublime'
+		];
+		foreach ( $editors as $name => $value ) {
+			echo '<option value="' . esc_attr( $value ) . '" ' . selected( $value, $editor ) . '>' . esc_html( $name ) . '</option>';
```
johnbillion: Can you check this use of `selected()`? Looks like the `$echo` parameter should be false.

rmccue: Ah, good catch; I didn't notice because browsers ignore non-`<option>` inside a `<select>` ¯\_(ツ)_/¯

rmccue:
@johnbillion Updated to fix that issue.

Also, forgot to include a screenshot earlier:
Screen Shot 2019-11-20 at 17 20 23

crstauf:
@rmccue To be clear, I'm in favor of localStorage as the primary setting given [the point](https://github.com/johnbillion/query-monitor/pull/479#issuecomment-555976224) you made. 

What I am wondering and am unclear on, as to the reasons to not have user meta act as a fallback to the localStorage value? It doesn't seem like a big hurdle to overcome (more than happy to work on adding that to the PR myself)...?

crstauf:
Or not. 😉

johnbillion:
Happy for the convo to carry on!

rmccue:
Thanks @johnbillion!

@crstauf IMO, it leads to a split-brain problem where updating the preference doesn't actually update your preference.

Let's say you set the editor to VS Code in Firefox. You now open Chrome, log in, and your editor is VS Code.

If you now switch the editor to Atom in Chrome, it only updates Chrome (and future browsers), it doesn't update Firefox. If you go back to Firefox and update your editor to phpStorm, confusingly it'll apply to Firefox and any new browsers, but not Chrome.

It seems like weirdly complex behaviour which saves a click or two at the expense of being much tougher to explain. IMO, it doesn't make sense to do both, it only makes sense to do one or the other.

crstauf:
@rmccue That makes total sense; thank you for explaining. Wish there was a way to reduce clicks all around, but definitely too complex.

crstauf:
Encountered this again, where the use of `user_has_cap` to always show QM output (specifically in new incognito windows) means having to set the editor cookie on every instance and then refresh.

[This comment from John](https://github.com/johnbillion/query-monitor/pull/479#issuecomment-543124239) mentions using the `qm/output/file_link_format` filter, and that works well. 

~IMO a constant would be better (John [preferred](https://github.com/johnbillion/query-monitor/pull/479#issuecomment-543124239) not using a constant), as that could be defined in the environment's `wp-config.php` file, but a filter is fine for now; just have to make sure it doesn't make it's way into version control.~

Turns out you can call `add_filter()` after `wp-settings.php` is included in `wp-config.php`: you have to disregard the "That's all, stop editing!" warning in `wp-config-sample.php`, but it works, and being in `wp-config.php` means it's environment-independent and not committed into version control. 👍

Issue #473, opened 2019-09-23
[Forward Compatibility] Deprecated jQuery


pentatonicfunk:
In case WP updated their JQuery Version / Themes using updated jQuery Version. these are deprecated:
- jQuery.fn.click() event shorthand is deprecated
- jQuery.fn.change() event shorthand is deprecated
- jQuery.parseJSON is deprecated; use JSON.parse

johnbillion:
Thanks!

absolutelynotjames:
Hey there, any chance I could assist with this? I've been looking to start making contributions to an open source project and this is one that I use almost every day. This issue looks pretty straightforward and I think I've got everything pulled down and set up in my local environment. I'd be happy to come back with a PR to tackle these!

Buzut:
Could it work without jQuery altogether? As the main usage of the plugin is to improve performance, on many custom themes, I just don't use jQuery and can't have the plugin to work on the frontend.

johnbillion:
Query Monitor works without jQuery, just with reduced functionality (check the Settings panel for the constant you can set to prevent QM from depending on jQuery). My long term aim is to remove jQuery entirely but it's not high priority.

Buzut:
Thank you for the answer (and sorry for the delay for reponding!).

PR #480, opened 2019-10-19
Enhancement/deprecated jquery

absolutelynotjames:
Addresses #473 

- Tested against jquery-migrate 3.1

- jquery-migrate still complains about $.unique(), I think because it's used internally by $.find() in the 1.x branch of jquery?

Please let me know if anything doesn't look quite right or if there's anything I could have done better!

Thanks!


johnbillion:
Thanks for the PR! I'll try to find some time over the next week or so to review.

johnbillion:
@absolutelynotjames Thanks again for the PR.

There seems to be a problem with your branch and I'm not sure exactly what it is. I think your branch is somehow detached from the main tree. I'll try to find some time to figure it out.

In the meantime: The second of these three commits adds a bunch of changes which will break QM because they're removing calls which trigger an event. Do you have some time to correct these? Let me know. The main correction is that they need to call `{element}.trigger( {event} )` rather than adding an event handler.

absolutelynotjames:
@johnbillion - Thanks for the feedback, sounds straightforward. I'll take a look and work to get these resolved!

absolutelynotjames:
Thanks again for flagging those mistakes in my code - now that I'm looking at it that was kind of a silly mistake. I changed all of the .on() methods to .trigger() as they should be. I did a run through test of all the button/select elements on my local dev copy and I think everything seems to work as expected.

I also tried to bring my branch up to date with yours, and included that merge in a squashed commit. I hope this helps, but I see there are a couple of merge conflicts.

Let me know if anything else looks off or if there's anything else I can do about my branch deviating from the main tree.

Thanks!

johnbillion:
Thanks again for those updates James. I'm not sure what's up with your branch, but it's completely detached from the main tree:



I think your best bet here is to start again from the develop branch and re-commit those changes without squashing or rebasing, otherwise you'll still have a detached branch.

PR #483, opened 2019-11-07
Implement client side rendering

johnbillion:
This starts us down the path of switching over Query Monitor's output to client-side rendered React components.

This has numerous benefits including:

* Better performance (no need to render a big pile of HTML)
* Allowing for state to be swapped out in order to show data for requests other than the current page load (eg. Ajax, REST API, POSTs before a redirect, and historical requests)
* Panels which can be reused outside of the context of WordPress, eg in loggers or browser extensions
* Data accessible elsewhere in the client, eg in the developer tools panel

There's still lots to do, but the concept has been proven.

See #291.

## Progress

* [x] Proof of concept
* [x] Functional
* [ ] Quality
* [ ] Functionally complete
* [ ] Shiny and ready to go

## Aims

* Feature parity with all existing panels.
* JSON data payload to hydrate the client-side panels. Remove unnecessary properties as we go.
* Retain support for server-side rendered panels so add-ons for QM and Debug Bar continue to work.
* ~~Partial client-side rendering. The main panel container and its menu tabs will remain server-side rendered for the time being. Individual panel contents will be converted to client-side rendering.~~
* Retain and continuously improve the existing server-side panel rendering until all client-side panels are complete.
* ~~Publish the UI components as an NPM package~~
* Remove dependency on jQuery

## Considerations

* Bundling for WP backwards compatibility. `wp.i18n` isn't bundled with WP prior to 5.0.
* Bundling in general. Should React be bundled or should core's be used?
* Using data across components, eg most stuff in the Overview panel
  - Slot & Fill?

## Usage

    npm install --legacy-peer-deps && npm run watch

## Per-panel Tasks

The existing collectors and outputters in QM are very well separated, but nevertheless as each panel is worked on the following actions may need to be made:

* Move logic out of the HTML outputter and into the corresponding collector.
* Move translatable strings out of the collector and into the corresponding HTML outputter and React component.

## Panel Progress

* [ ] `overview`
   - Use a slot that other components can fill with their data
* [x] `admin`
* [x] `assets_scripts`
* [x] `assets_styles`
* [x] `block_editor`
* [ ] `cache`
* [x] `caps`
* [x] `conditionals`
* [x] `db_callers`
* [x] `db_components`
* [x] `db_dupes`
* [x] `db_queries`
* [x] `doing_it_wrong`
* [x] `environment`
* [ ] `headers`
* [x] `hooks`
* [x] `http`
* [x] `languages`
* [x] `logger`
* [x] `multisite`
* [x] `php_errors`
* [ ] `raw_request`
* [x] `request`
* [x] `theme`
* [x] `timing`
* [x] `transients`
* [ ] Hooks in Use panels
* [x] Settings panel

## UI Component Progress

* [x] Toggles
* [x] Filter dropdowns
* [ ] Filter links
* [ ] Table sorters
* [x] Clickable stack traces
* [x] Table footers

## Nope

The following panels will remain server-side rendered:

* [x] `debug_bar`


output/html/assets_styles.js
```diff
@@ -0,0 +1,20 @@
+import React, { Component } from 'react';
+import Assets from '../assets.js';
+import Tabular from '../tabular.js';
+import { __, _x, _n, sprintf } from '@wordpress/i18n';
+
+class Styles extends Component {
```
tomjn: Does it make sense for there to be a Scripts and a Styles component with exactly the same implementation?

johnbillion: Currently I'm building the components to match the existing PHP implementations. You're right that with the new model this is probably not needed, but then I haven't fixed up all the labelling yet so we'll see.

jordantrizz:
Just wanted to say thank you for working on this!

jordantrizz:
Just wondering if this is functional yet?

johnbillion:
For anyone following along: #1068.

PR #504, opened 2020-03-02
Add new qm/collect/silent_http_error_statuses filter to silence specific HTTP API requests that result in 400-500 HTTP status codes

pbiron:
this PR addresses #474 in a slightly different way from my previous PR.

### DIFF
--- a/collectors/http.php
+++ b/collectors/http.php
@@ -211,6 +211,33 @@ class QM_Collector_HTTP extends QM_Collector {
 		$this->data['http'][ $args['_qm_key'] ]['transport'] = $this->transport;
 		$this->info      = null;
 		$this->transport = null;
+
+		/**
+		 * Filters the HTTP response statuses that should be silenced.
+		 *
+		 * @since 3.5.3
+		 *
+		 * @param int[] $silenced HTTP response statuses to be silenced.  Default: empty array.
+		 * @param array $http {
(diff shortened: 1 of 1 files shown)


pbiron:
It applies the filter in `QM_Collector_HTTP::log_http_response()` and adds the results of the filter
to a new member of the HTTP "object" that QM stores.  Then in `QM_Collector_HTTP::process()` it checks that member to decide whether the request was an "error".

This allows plugins to bracket their calls to `wp_remote_request()` with `add_filter()`/`remove_filter()`, instead of adding the filter "globally" and then in their callback having to check the backtrace.

I've been using this approach in every version of QM since 3.4.0 (i.e., I've been applying this patch to my local copy of QM) and have not noticed it adversely affect QM.

pbiron:
The wordpress-beta-tester plugin already hooks into the filter in this PR (see https://github.com/afragen/wordpress-beta-tester/blob/develop/src/WPBT/WPBT_Beta_RC.php#L255).   Of course, we can always change
that part of beta tester if, for example, you have a better way of achieving this goal.

pbiron:
Here is a screenshot of beta tester without the PR applied.
qm-wp-beta-tester-without-pr-applied

And here is a screenshot with it applied.
qm-wp-beta-tester-with-pr-applied

johnbillion:
Closing as per #474

PR #505, opened 2020-03-03
Fix #498 Stack traces showing at all times

### LINKED ISSUE #498: Stack traces showing at all times
Hi,

We recently upgraded Query Monitor to the latest version (3.5.2) and noticed that when we had an uncaught error, we were now displaying a stack trace in production even thought WP_DEBUG is set to false. This was a bit of a surprise. 

When we reverted back to our previous version 3.3.0 we got the expected behaviour of a message stating: 'There has been a critical error on your website.' 

We narrowed down the issue to the QM_Collector_PHP_Errors class around [here](https://github.com/johnbillion/query-monitor/blob/develop/collectors/php_errors.php#L57). In the output_fatal code it echos out to the screen the stack trace. 

We expected this behaviour to be controlled by WP_DEBUG. We completely disabled PHP errors by using the follow snippet:

    error_reporting(0);
    @ini_set('display_errors', 0);

And the stack trace still appeared which caused a bit of confusion. 

Hope this is a clear issue report, if not then please let me know.

Thanks,
Stephen

glagonikas:
Check if display_errors is enabled before showing stack trace. Fixes #498

### DIFF
--- a/collectors/php_errors.php
+++ b/collectors/php_errors.php
@@ -47,6 +47,10 @@ class QM_Collector_PHP_Errors extends QM_Collector {
 	public function exception_handler( $e ) {
 		require_once __DIR__ . '/../output/Html.php';
 
+		if ( empty( $this->display_errors ) ) {
+			return;
+		}
+
 		if ( is_a( $e, 'Exception' ) ) {
 			$error = 'Uncaught Exception';
 		} else {


johnbillion:
Thanks for the PR @glagonikas ! I won't have time to look at this for a couple of weeks, but I'll get  to it just as soon as I can.

glagonikas:
Hi, @johnbillion I was wondering if you have time to look into this? It's quite important for this bug to be resolved quickly as it's exposing sensitive information when the plugin is active

PR #507, opened 2020-03-15
Fix PHPDoc

szepeviktor:
Discovered by @phpstan

### DIFF
--- a/classes/Plugin.php
+++ b/classes/Plugin.php
@@ -21,7 +21,7 @@ abstract class QM_Plugin {
 	/**
 	 * Returns the URL for for a file/dir within this plugin.
 	 *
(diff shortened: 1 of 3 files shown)


johnbillion:
Thanks.

What's the rule about `$this` as a parameter name for filters? I understand the reasoning (so implementations don't try to use a function parameter called `$this`) but it does mean the docblock parameter has a different name to the actual parameter.

szepeviktor:
> it does mean the docblock parameter has a different name to the actual parameter.

Yes, `array()` is also named `$colletors`.
Docblocks list the **name** of the parameter, not its value.

szepeviktor:
BTW To benefit from PHPStan you have to pay some little price.
But at the end paying these little prices end up to be beneficial too.

szepeviktor:
Thank you.

PR #522, opened 2020-05-29
Support passing backtrace to QM_Backtrace

tillkruss:
This PR adds support for constructing a backtrace with a previously collected backtrace.

My particular use case is related to object caching and decoupling logs from the QM plugin.

### DIFF
--- a/classes/Backtrace.php
+++ b/classes/Backtrace.php
@@ -54,8 +54,8 @@ class QM_Backtrace {
 	protected $calling_line    = 0;
 	protected $calling_file    = '';
 
-	public function __construct( array $args = array() ) {
-		$this->trace = debug_backtrace( false );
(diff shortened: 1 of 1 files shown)


classes/Backtrace.php
```diff
@@ -54,8 +54,8 @@ class QM_Backtrace {
 	protected $calling_line    = 0;
 	protected $calling_file    = '';
 
-	public function __construct( array $args = array() ) {
-		$this->trace = debug_backtrace( false );
+	public function __construct( array $args = array(), array $trace = null ) {
+		$this->trace = $trace ? $trace : debug_backtrace( false );
```
szepeviktor, line 58:
```suggestion
		$this->trace = null === $trace ? debug_backtrace( false ) : $trace;
```

tillkruss: Thanks, resolved.

PR #525, opened 2020-06-19
add TextMate to list of supported editors

paulschreiber:


output/Html.php
```diff
@@ -459,6 +459,8 @@ public static function get_editor_file_link_format( $editor, $default_format ) {
 				return 'atom://open/?url=file://%f&line=%l';
 			case 'sublime':
 				return 'subl://open/?url=file://%f&line=%l';
+			case 'textnate':
```
johnbillion, line 462: Typo here.

PR #527, opened 2020-06-21
Add Plugin information (incl. execution time)

NilsHoyer:
Since two years, I miss a feature in Query Monitor: An overview of the plugins and their execution time.

It list all active plugins with the entiered execution time and a detailed list of each hook for all plugins.

Now, I implemented it and I hope you love too.

### DIFF
--- a/classes/Hook.php
+++ b/classes/Hook.php
@@ -21,7 +21,13 @@ class QM_Hook {
 
 				foreach ( $callbacks as $callback ) {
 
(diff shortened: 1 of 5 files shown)


lkraav:
Interesting. Mind posting a screenshot of what the expected display result should be?

NilsHoyer:
I found a mistake in my thinking and found out, my solution does not work.

PR #529, opened 2020-07-10
Fix errors caused by missing user or group IDs

### LINKED ISSUE #514: "array offset on value of type bool" notice with PHP 7.4
The following notice is triggered on every page load in PHP 7.4:

> PHP Notice:  Trying to access array offset on value of type bool in wp-content/plugins/query-monitor/collectors/environment.php on line 265

morganestes:
Although `posix_getpwuid()` and `posix_getgrgid()` should return arrays, they sometimes return `false` if there a problem reading the user/group info from the filesystem. To avoid the "can't access index of a bool" error, check to make sure the results are truthy, and skip setting the user until we have better values later on.

Fixes #514.


johnbillion:
Thanks!

PR #531, opened 2020-07-23
Disable db.php symlink

crstauf:
I run QM conditionally on production (by IP address), but the `db.php` file is put in place, meaning that all queries are run through QM's drop-in when not necessary. This PR introduced a constant to prevent symlinking, and updates messaging in various places.

If approved:
- [ ] Update wiki with information about constant.


crstauf:
@johnbillion Thanks for merging this!

Checking if the wiki was updated with info about the constant?

johnbillion:
It's on my list. Got a deployment issue to look at first.

PR #549, opened 2020-10-14
Changes dispatch action priority from lowest to highest

pschoffer:
This enables us to catch also queries that happen during shutdown
actions. While in the current implementation these would happen only AFTER the dispatch.

Looking at the git history, there seems to be no indication of why the 0 priority (the first action to run) was chosen.

@johnbillion can you recall what was the reasoning behind that? Or do you have any concerns about changing the priority?

### DIFF
--- a/dispatchers/AJAX.php
+++ b/dispatchers/AJAX.php
@@ -21,7 +21,7 @@ class QM_Dispatcher_AJAX extends QM_Dispatcher {
 	public function __construct( QM_Plugin $qm ) {
 		parent::__construct( $qm );
 
-		add_action( 'shutdown', array( $this, 'dispatch' ), 0 );
+		add_action( 'shutdown', array( $this, 'dispatch' ), PHP_INT_MAX );
 
 	}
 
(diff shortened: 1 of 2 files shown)


johnbillion:
There is a reason for the current priority of `0`, I just need to remember it. 😬

pschoffer:
@johnbillion I hate to push you on this, but it would be great to get more info on this Any input would be a great help!

johnbillion:
I found the reason for this priority of `0`: 52c50f0.

This change was made waaayyy back when Query Monitor didn't have the concept of separate dispatchers for HTML, Ajax, Redirects, etc. Priority `0` is used because [WP core hooks in on priority `1` and flushes all the output buffers](https://github.com/WordPress/wordpress-develop/blob/0a3a3c5119897c6d551a42ae9b5dbfa4f576f2c9/src/wp-includes/default-filters.php#L382) which means QM couldn't output headers for Ajax responses if it ran after this.

As separate dispatchers for each context now, the HTML dispatcher could indeed run on a later priority in order to collect data for processing that happens inside the `shutdown` hook, but the Ajax dispatcher needs to remain on `0`.

I'll check what else happens on `shutdown`. QM will still need to run before things get destroyed.

Issue #539, opened 2020-09-23
Iteration Plan: Autumn/Winter 2020


johnbillion:
This plan captures the current high level goals for Query Monitor. The list might change over time and the next plan will start in Spring 2021.

This plan isn't tied to any particular releases. Some planned items need an issue to be opened as they've been moved here from my notes app. Other bug fixes will be included in releases but aren't necessarily listed here.

## Priority Fixes

* [ ] Make `QM_Backtrace` more resilient, reliable, add tests - #564
  - [ ] #495
  - [ ] #494 
  - [ ] #461 
  - [ ] #433 
  - [ ] #340 
* [x] Remove warning when certification verification is disabled for local URLs #537 

## New Panels

* [ ] `WP_Error` object tracking #559
* [ ] Object cache statistics standard #518

## General Features

* [x] Always show the Logs panel, with a link to help docs
* [x] Remove QM by default from places where it technically should show but practically isn't useful
  - [x] #355
  - [x] #335
* [ ] 404 helper #342
* [ ] ~~Ability for code partway through a request to prevent QM's output~~ This is already possible with the `qm/dispatch/html` filter. Need a way to cease the data collection and processing though.
* [ ] Add time on running hooks #260 (stretch goal?)

## Performance and Dependencies

* [x] Prepare for jQuery updates #473
* [ ] Limit the number of HTTP headers for PHP errors that occur during Ajax, REST API, etc #275 

## WordPress Core Tasks

* [X] Add `WP_Error` tracking [#WP40568](https://core.trac.wordpress.org/ticket/40568)
* [ ] Add an action when a rewrite rule is added or removed, including endpoints etc

## Meta

* [x] GitHub action for building a tag so the Composer package can be un-abandoned #551 
* [x] Switch to GitHub action for semver milestone management (currently Probot) #552
* [ ] General website updates, TBD
* [ ] Publish info about [sponsorship](https://github.com/sponsors/johnbillion)

## Ongoing

Long (long) term work which is in progress but very unlikely to ship soon

* [ ] Implement client side rendering 🚀  #483 , the God-ticket which will ultimately fix or unblock:
  - #469 
  - #456 
  - #379 
  - #354 
  - #291 
  - #223 
  - #174 
  - #173 
  - #39 
* [ ] Hook inspection
* [ ] Component insights
* [ ] Flow explorer
* [ ] Document all the actions and filters

---

[...]

PR #557, opened 2020-11-09
Support backtrace filtering by instance

tillkruss:
I need to filter some classes, methods and functions, but only for a single QM view, not all of them.

### DIFF
--- a/classes/Backtrace.php
+++ b/classes/Backtrace.php
@@ -283,7 +283,6 @@ class QM_Backtrace {
 		}
 
 		$return = $frame;
-		
 		$ignore_class = array_merge( self::$ignore_class, $this->args['ignore_class'] );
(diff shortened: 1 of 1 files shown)


johnbillion:
In principle yes this is needed. Thanks for the PR!

I started some work on this a while ago but didn't get far. Support for ignoring given hook names is needed too so I can remove all the unreliable `ignore_frames` and `ignore_current_filter` calls, which rely on the stack trace being somewhat known for any given call.

This fits in nicely with my aims in #539 to make the stack traces more reliable.

johnbillion:
I'm currently working on a test framework for `QM_Backtrace` so you can pass it a stack trace and the expected caller list and it asserts they match. I'll open a PR in the next few days.

PR #585, opened 2021-04-16
Fix notice if caller isnt an array

kkmuffme:


johnbillion:
Thanks for the PR! Do you have some info about how to reproduce this problem please?

kkmuffme:
Not exactly after 1.5 years.
From what I can see in the code it happens if reset() in get_caller() returns false - and reset() returns false if the array is empty. The trace array is empty when the backtrace is empty.

When is the backtrace empty? e.g. when we are inside of a callback of a `register_shutdown_function`
I wouldn't have PRed this if that wasn't a common enough notice/issue back then though.

PR #588, opened 2021-04-26
Fix highlighting of TH cells

crstauf:
Fixes `.qm-highlight` class not applying highlight style to `th` table cells.


johnbillion (approved):

johnbillion:
This looks good but for some reason this commit is on a detached tree and can't be merged. Want to try again with a new branch from develop and a new PR?

johnbillion:
Scrap that, it seems to be a problem with my local working copy. I'll sort it.

PR #631, opened 2021-12-27
Add `qm/component_type/{$type}` filter

tillkruss:
When using symlinked (mu-)plugins, there is currently no way to filter the component name, context and type.

The `qm/component_dirs` filter is called very early and the result is cached without a plugin able to modify this later on.

The only other approach would be to apply the `qm/component_dirs` for every `get_file_dirs()` call.

### DIFF
--- a/classes/Util.php
+++ b/classes/Util.php
@@ -252,7 +252,7 @@ class QM_Util {
 			case 'unknown':
 			default:
 				$name = __( 'Unknown', 'query-monitor' );
-				
+
 				/**
 				 * Filters the type of a custom or unknown component.
 				 *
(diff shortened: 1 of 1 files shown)


johnbillion:
Thanks. What's the use case for changing `$type` and what do you change it to? Do you just use it to affect the names of the two subsequent filters?

tillkruss:
@johnbillion: Object Cache Pro is often installed as a symlinked must-use plugin and when it throws let's say a deprecated notice the path/filename of the error points somewhere outside the paths returned from `get_file_dirs()` and always ends up being `Unknown`.

Even when using the `qm/component_name/unknown` filter to fix the name, because of it's `unknown` type QM will show either "WordPress Core" or whatever plugin registers an error handler.

As mentioned, we could also run the `qm/component_dirs` for every `get_file_dirs()` call instead of introducing this filter.

Side note, I ran into this because `get_file_dirs()` get called immediately on PHP 8.1, because QM itself throws some deprecation notices.

tillkruss:
Thanks! 🥰

PR #632, opened 2021-12-27
Introduce plugin version constant

tillkruss:
It's rather tedious to read the current QM version, since it may be installed as a MU-plugin as well, so relying on `query-monitor/query-monitor.php` isn't an option.

Updating the constant obviously need to be part of the version release process.

```php
$qmVersion = null;
$qm = QueryMonitor::init();

if (isset($qm->file) && is_readable($qm->file)) {
    $qm = get_file_data($qm->file, false, false);
    $qmVersion = $qm['Version'];
}
```

### DIFF
--- a/query-monitor.php
+++ b/query-monitor.php
@@ -33,6 +33,8 @@ if ( ! defined( 'ABSPATH' ) ) {
 	exit;
 }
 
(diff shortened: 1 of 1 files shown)


johnbillion:
Thanks. No change needed to the release process as the existing version bump script handles this.

Issue #638, opened 2022-01-02
QM_Util::get_query_type method fails with certain mysql comment types


braders:
I recently came accross a case where the following notice was originating from within Query Monitor:

`Notice: Undefined variable: type` in `wp-content/plugins/query-monitor/classes/Util.php:393`

It turns out that this was being caused by the following Mysql query:

```/*!40101 SET NAMES utf8 */```

This is using a style of comment specific to MySql which means that the command (in this case `SET NAMES utf8`) should only run if the MySql server vesion is greater than `4.1.1`. See https://dev.mysql.com/doc/refman/8.0/en/comments.html for more details on this syntax.

Query monitor is stripping out all comments before determining the type, and this means that `$sql` is an empty string at the point the `$words` regex is run:

```
if ( 0 === strpos( $sql, '/*' ) ) {
  // Strip out leading comments such as `/*NO_SELECT_FOUND_ROWS*/` before calculating the query type
  $sql = preg_replace( '|^/\*[^\*/]+\*/|', '', $sql );
}
```

I see a couple of options:

**Option 1 (easier)**

Check that the `$words` array is not empty before reading `$words[0]`. If it is empty fallback to `Unknown`. So something like this:

```
$words = preg_split( '/\b/', trim( $sql ), 2, PREG_SPLIT_NO_EMPTY );
$type = __( 'Unknown', 'query-monitor' );
if ( isset( $words[0] ) ) {
$type = strtoupper( $words[0] );
}
```

**Option 2**

Update the comments regex to ignore MySql-specific comments (i.e. `/*!` rather than `/*`), and instead read the first word preceeding the comment. So for the query `/*!40101 SET NAMES utf8 */`, the comment `/*!40101` would be ignored, and the word `SET` would be returned.

---

Depending on which option the maintainers of this plugin determine is best, I may be able to open a pull request to resolve. I appreciate this is rare MySql syntax (and the query above is not one I wrote!) and therefore I think I'd be inclined to lean towards option 1.

johnbillion:
Thanks for the detailed report! Option 1 sounds sensible for now and I can look at implementing option 2 as a later enhancement.

PR #639, opened 2022-01-03
Handle unexpected SQL in QM_Util::get_query_type method

braders:
Resolves https://github.com/johnbillion/query-monitor/issues/638

Checks that `$words[0]` exists, and if not fallsback to the string "Unknown".


johnbillion (approved):
Thanks for the PR. I've also adjusted the output so any "Unknown" query types don't get formatted with line breaks, syntax highlighting, etc.

PR #664, opened 2022-03-15
Allow Control of QM_ERROR_FATALS

Kevinlearynet:
PHP 8+ will always show _Deprecated_ logs under the "Logs" panel, regardless of the value of `error_reporting()`. If you don't want these there isn't currently a universal way to remove them, they appear even when the "Environment" panel shows `E_DEPRECATED` as disabled.

Allowing control of the `QM_ERROR_FATALS` constant provides a simple solution.

### DIFF
--- a/collectors/php_errors.php
+++ b/collectors/php_errors.php
@@ -6,10 +6,9 @@ if (! defined('ABSPATH')) {
   exit;
 }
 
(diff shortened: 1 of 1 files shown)


johnbillion:
Thanks for the PR!

PR #687, opened 2022-08-05
Update theme.php

abdihaikal:


johnbillion:
Can you provide some information about why this change is needed?

PR #693, opened 2022-09-17
Use unit symbol for seconds

IanDelMar:
This PR replaces the translatable string 'S' (for seconds) in the admin bar with the [SI](https://www.bipm.org/documents/20126/41483022/SI-Brochure-9-EN.pdf/2d2b50bf-f2b4-9661-f402-5f9d66e4b507?version=1.10&t=1632138671693) unit symbol (s) that does not have a plural form. Motivation: "floating-point values usually represent physical entities for which singular and plural don’t clearly apply." - [GNU gettext manual](https://www.gnu.org/software/gettext/manual/gettext.html#Additional-functions-for-plural-forms)

By also removing the space between the value and the unit symbol and using the existing translatable string _x( '%ss', 'Time in seconds', 'query-monitor' ) one can avoid the need for new translations. The space gets stripped anyway [here](https://github.com/johnbillion/query-monitor/blob/b477463fcc26ad8f1ea410484f340e7f589a0f6a/output/html/db_queries.php#L561) and [here](https://github.com/johnbillion/query-monitor/blob/b477463fcc26ad8f1ea410484f340e7f589a0f6a/output/html/overview.php#L399).

This also fixes the PHP 8.1 deprecation notice "Implicit conversion from float [some floating-point number] to int loses precision in wp-includes/pomo/plural-forms.php on line 243" and on lines 246 and 247 that originates from https://github.com/johnbillion/query-monitor/blob/b477463fcc26ad8f1ea410484f340e7f589a0f6a/output/html/db_queries.php#L516

I found no other instance in the code where uppercase S is used to denote seconds, it's always lowercase s.

There is also a space in `'%s MB'` and in `'%s Q'` that gets stripped. I left those strings unchanged so existing translations still apply. If at some point the spaces are removed, I'd also go for using `'%dQ'` [here](https://github.com/johnbillion/query-monitor/blob/b477463fcc26ad8f1ea410484f340e7f589a0f6a/output/html/db_queries.php#L531-L537) and [here](https://github.com/johnbillion/query-monitor/blob/b477463fcc26ad8f1ea410484f340e7f589a0f6a/output/html/db_queries.php#L546-L552).

### DIFF
--- a/output/html/db_queries.php
+++ b/output/html/db_queries.php
@@ -520,7 +520,7 @@ class QM_Output_Html_DB_Queries extends QM_Output_Html {
 					number_format_i18n( $db->total_time, 2 )
 				);
 
-				/* translators: %s: Number of database queries. Note the space between value and unit. */
+				/* translators: %s: Number of database queries. Note the space between value and unit symbol. */
 				$text = _n( '%s Q', '%s Q', $db->total_qs, 'query-monitor' );
 
 				// Avoid a potentially blank translation for the plural form.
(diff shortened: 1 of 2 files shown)


johnbillion:
@IanDelMar Thanks. The odd formatting of units in the admin toolbar is mostly legacy. I've pushed a couple of adjustments so the `<small>` tag is only output when it's needed.

PR #694, opened 2022-09-27
Count array elements outside the foreach loop

IanDelMar:
This PR moves the counting of array elements outside the `foreach` loop.

Motivation: The loop does not alter the array, therefor the number of array elements is constant and it's sufficient to count elements once.


johnbillion:
Thanks 👍

IanDelMar:
Is it `foreach ( $data->dbs as $key => $db ) {` instead of `foreach ( $data->dbs as $db ) {` on purpose? `$key` is never used.

PR #699, opened 2022-10-18
Update minimum required WP version.

peterwilsoncc:
The function `get_language_attributes()` is unavailable prior to version 4.3.

This causes a fatal error on previous version of WP.

Screen Shot 2022-10-19 at 9 41 00 am

Screenshot from WP 4.2


johnbillion:
Cheers @peterwilsoncc! I think I'll wrap this call in a conditional check instead so the minimum supported version doesn't need to be bumped (yet).

PR #702, opened 2022-10-21
Use composer v2 for release builds

tomjn:
By specifying the `composer:v2` tool in the setup-php action we can ensure composer v2 and not v1 is used, which should avoid the PHP warning with some sites in the current release

### DIFF
--- a/.github/workflows/acceptance-tests.yml
+++ b/.github/workflows/acceptance-tests.yml
@@ -56,6 +56,7 @@ jobs:
       uses: shivammathur/setup-php@2.7.0
       with:
         php-version: 8.0
+        tools: composer:v2
         coverage: none
       env:
         fail-fast: true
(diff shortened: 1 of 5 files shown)


tomjn:
I don't know, but I do know that currently this causes issues for all users installing QM via composer 2

tomjn:
This line implies it does run on 5.6:

> On PHP 5.6+, the class map is also cached in opcache which improves the initialization time greatly. If you make sure opcache is enabled, then the class map should load almost instantly and then class loading is fast.

https://getcomposer.org/doc/articles/autoloader-optimization.md

Though the implication is that if it did not support 5.6 then anybody running v3.9 of QM would encounter issues that they didn't report here. it's also possible the action used might change to v2 in the future by default anyway so you'd want to be specific.

You can also enforce 5.6 compatibility via `composer.json` but I see it's added via `"php": ">=7.2.0",` at the moment.

You can also specify it via:

```json
"config": {
    "platform": {
        "php": "5.6"
    }
}
```

This should make sure the packages it chooses are based on a PHP 5.6, even if you're running composer under PHP 8.1

johnbillion:
Yeah the master branch is already on 7.2 in preparation for QM 3.11.

PR #706, opened 2022-11-10
Sync number of accepted args with the callback's expected args

ocean90:
Fixes a phpstan error:

Error: Callback expects 2 parameters, $accepted_args is set to 4.


johnbillion (approved):

johnbillion:
Thanks for the PRs 👍 . I need to pin some of the test dependencies so they're more stable and don't cause failures in PRs that cause them to get updated.

PR #710, opened 2022-11-22
language collector 'file' and 'line' sometimes missing

PhilMakower:
in log_file_load() the $caller doesn't always have 'file' or 'line', I think when the function is filtered, or when we are in a hook.
In these cases, it has 'calling_file' and 'calling_line' instead, so we can copy those values to 'file' and 'line'.


johnbillion:
Thanks very much for the PR @PhilMakower but I've found that the root cause of this is in the `QM_Collector` class. I'll get it fixed in the next release. Cheers!

PR #717, opened 2023-01-16
Try a more reliable approach to waiting for Selenium and the database during tests

johnbillion:
The acceptance tests, and to a lesser degree the integration tests, are flakey during the CI runs because sometimes the Chrome container or the MariaDB container isn't ready when the tests start.

The workflows sleep for 10 seconds after starting the containers in an attempt to mitigate this, but that's not ideal and not reliable.

This PR attempts to correctly wait until the corresponding service is fully ready and healthy before running the tests

### DIFF
--- a/bin/acceptance-tests.sh
+++ b/bin/acceptance-tests.sh
@@ -6,9 +6,9 @@ set -eo pipefail
 
 # Prep:
 docker-compose --profile acceptance-tests up -d
-WP_PORT=`docker port qm-server | grep "[0-9]+$" -ohE | head -1`
-CHROME_PORT=`docker port qm-chrome | grep "[0-9]+$" -ohE | head -1`
-DATABASE_PORT=`docker port qm-database | grep "[0-9]+$" -ohE | head -1`
+WP_PORT=`docker inspect --type=container --format='{{(index .NetworkSettings.Ports "80/tcp" 0).HostPort}}' qm-server`
+CHROME_PORT=`docker inspect --type=container --format='{{(index .NetworkSettings.Ports "4444/tcp" 0).HostPort}}' qm-chrome`
+DATABASE_PORT=`docker inspect --type=container --format='{{(index .NetworkSettings.Ports "3306/tcp" 0).HostPort}}' qm-database`
 WP_URL="http://host.docker.internal:${WP_PORT}"
 WP="docker-compose run --rm wpcli --url=${WP_URL}"
 
--- a/bin/acceptance-tests.sh
+++ b/bin/acceptance-tests.sh
@@ -12,11 +12,16 @@ DATABASE_PORT=`docker inspect --type=container --format='{{(index .NetworkSettin
 WP_URL="http://host.docker.internal:${WP_PORT}"
 WP="docker-compose run --rm wpcli --url=${WP_URL}"
 
-# Wait for the web server and the database:
+# Wait for the web server:
 ./node_modules/.bin/wait-port -t 10000 $WP_PORT
-./node_modules/.bin/wait-port -t 10000 $DATABASE_PORT
 
-# Wait for Selenium
+# Wait for MariaDB:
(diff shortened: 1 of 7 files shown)


szepeviktor:
`docker port qm-server | grep "[0-9]+$" -ohE | head -1`
These port number parsers look too _futuristic_.
Docker must have JSON output ...

szepeviktor:
```bash
docker inspect --type=container qm-server --format='{{(index .NetworkSettings.Ports "80/tcp" 0).HostPort}}'
```
The output is 32768 🚀

bin/acceptance-tests.sh
```diff
@@ -6,12 +6,17 @@ set -eo pipefail
...
+WP_PORT=`docker inspect --type=container --format='{{(index .NetworkSettings.Ports "80/tcp" 0).HostPort}}' qm-server`
+CHROME_PORT=`docker inspect --type=container --format='{{(index .NetworkSettings.Ports "4444/tcp" 0).HostPort}}' qm-chrome`
+DATABASE_PORT=`docker inspect --type=container --format='{{(index .NetworkSettings.Ports "3306/tcp" 0).HostPort}}' qm-database`
 WP_URL="http://host.docker.internal:${WP_PORT}"
 WP="docker-compose run --rm wpcli --url=${WP_URL}"
 
+# Wait for Selenium, the web server, and the database:
+./node_modules/.bin/wait-port -t 10000 $CHROME_PORT
+./node_modules/.bin/wait-port -t 10000 $WP_PORT
+./node_modules/.bin/wait-port -t 10000 $DATABASE_PORT
```
szepeviktor, line 18: @johnbillion Is it likely that Selenium opens its port first?
Aren't the other two faster to boot?

johnbillion: Yes probably but in my local testing there's barely any difference. The problem seems to be more with whether Selenium is actually ready, rather than whether its port is responding. I need to investigate it further.

szepeviktor:
There is Docker health check for you.
https://docs.docker.com/compose/compose-file/#healthcheck
After adding it to each container you can wait for them with a simple
`timeout 10 bash -c 'while [ "$(docker inspect ...)" != healthy ]; do sleep 1; done'`

szepeviktor:
The image author also talks about this started but not running problem.
https://github.com/seleniumhq-community/docker-seleniarm#using-a-bash-script-to-wait-for-the-grid
Selenium is written in Java, and a hello world in Java takes long seconds to execute.

Would you try the above shell script or the built-in `/opt/bin/check-grid.sh`? The previous section talks about is in their README.

johnbillion:
Well that seems to have worked nicely, thanks. Just need to do the same thing with MariaDB now.

szepeviktor:
> the same thing with MariaDB

It goes like `mysqladmin ping`.

szepeviktor:
PHP-FPM has a built-in status page.
https://www.php.net/manual/en/fpm.status.php
It can speak JSON. You may check whther `jq -r '."process manager"' equals `dynamic`.

szepeviktor:
as far as I know -u needs a space but -p does not

actually in scripts I prefer the long form for readability `--user=name --password=pass`

szepeviktor:
Waaaait! `-it` means you need a terminal.
Looking up docs which letter it is ... 🏃🏻‍♂️

szepeviktor:
The  `t`
```
  -t, --tty                  Allocate a pseudo-TTY
```

szepeviktor:
I think we should drop `i` too.
```
  -i, --interactive          Keep STDIN open even if not attached
```

szepeviktor:
Party time 🎉 John!

johnbillion:
Thanks for the pointers Viktor. I shall now retire as a programmer and never use a computer again.

johnbillion:
This isn't working reliably, the integration tests are still failing quite frequently because the database connection isn't ready. Ugh.

szepeviktor:
```bash
while ! curl -sSL "http://localhost:${CHROME_PORT}/wd/hub/status" 2>&1 | grep '"ready": true' >/dev/null; do
```

Do you think Seleniuim is not ready enough?

PR #718, opened 2023-01-21
Stop using Composer from GHA image

szepeviktor:
`shivammathur/setup-php` was later than Composer execution resulting usage of Composer from the GHA image.


szepeviktor:
@johnbillion Is this the ❌ we have been chasing in #717?

There is 1×"Waiting for MariaDB..."

PR #726, opened 2023-02-17
Remove substring selector for filtering

rebeccahum:
👋 
We've extended Query Monitor and are using `build_filter()`. However, we are seeing that the substring selector for filtering will match groups that share the same string.

For example, in filtering `WP_Object_Cache`, you would see:
```
WP_Object_Cache
WP_Object_Cache_global
```

Screenshot 2023-02-17 at 2 02 56 PM

This PR fixes that:
Screenshot 2023-02-17 at 2 05 22 PM

Uncertain though if there's a use case to select the substring.

### DIFF
--- a/assets/query-monitor.js
+++ b/assets/query-monitor.js
@@ -244,9 +244,9 @@ if ( window.jQuery ) {
 
 			if ( $(this).val() !== '' ) {
 				if ( hilite ) {
-					tr.filter('[data-qm-' + hilite + '*="' + val + '"]').addClass('qm-highlight');
+					tr.filter('[data-qm-' + hilite + '="' + val + '"]').addClass('qm-highlight');
 				}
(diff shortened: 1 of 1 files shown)


crstauf:
Related #223.

crstauf:
I do not believe this will work, because there are rows that have multiple filter keywords (ex: the `mu_plugin_loaded` action has `mu`, `plugin`, `loaded`). You'll probably get along further by writing your own filtering functionality.

johnbillion:
Thanks for the PR Rebecca.

Yes Caleb is correct, this change will break filters for rows that include multiple values for a given filter, which is the reason that substring selector is there.

I'll close this off as it's an edge case and not something I'm planning on fixing until I've switched over to client side rendering for the panels, then we can get rid of all the DOM based juggling.

Cheers!

PR #731, opened 2023-02-21
Change orphaned label to a span

joedolson:
Label element has no meaning without being associated with a form field. Buttons are self-labeling through their internal content or aria-label, and don't need a label associated.

### DIFF
--- a/output/Html.php
+++ b/output/Html.php
@@ -367,7 +367,7 @@ abstract class QM_Output_Html extends QM_Output {
 	 */
 	protected function build_sorter( $heading = '' ) {
 		$out = '';
-		$out .= '<label class="qm-th">';
+		$out .= '<span class="qm-th">';
 		$out .= '<span class="qm-sort-heading">';
 
 		if ( '#' === $heading ) {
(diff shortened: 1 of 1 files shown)


lkraav:
> Label element has no meaning without being associated with a form field.

I've found nothing inherently wrong with using `<label>` outside of form context. It certainly clarifies intent much cleaner than random descendant inventions like `<span class="label">`.

> Buttons are self-labeling through their internal content or aria-label, and don't need a label associated.

https://developer.mozilla.org/en-US/docs/Web/HTML/Element/label#interactive_content also recommends against placing interactive content inside `<label>`.

joedolson:
Using `label` outside of a form context doesn't have any real impact on real-world usage, either positive or negative. It doesn't clarify intent in any meaningful way, because its semantics are connective (binding fields to name). In this usage, where the sole usage of the container is in providing a CSS hook, there's no benefit to using label.

johnbillion:
The original intention of this label was that the whole column header should be clickable to sort it, but that's handled by the JavaScript handler for the header cell, so we're all good.

johnbillion:
Thanks for the PRs @joedolson! 👍

Issue #713, opened 2022-12-17
Accessibility: "Toggle more information" buttons


GaryJones:
Feedback given by @rianrietveld in https://github.com/johnbillion/query-monitor/pull/285#issuecomment-346938513 indicated that when buttons are extracted from their context (such as screen reader software can do), then they are not uniquely identifiable.

![Image showing VoiceOver with a load of repeated button texts](https://user-images.githubusercontent.com/2351479/33230529-8a66f352-d1e5-11e7-8c2a-bc90207a35b3.png)
(Rian's image from 2017 - the specific button texts have changed now, but the point still stands)

The buttons currently use the `aria-label` approach, so instead of a fixed `aria-label="Toggle more information"` attribute, it should include a reference to the thing is will display more details about. For example:
- `aria-label="Toggle more information about VaultPress_IXR_SSL_Client->query() caller"`, or 
- `aria-label="Toggle more information about Jetpack->add_configure_hook() hook callback"`.

johnbillion:
Fixed in #732 and #1063.

rianrietveld:
🎉

PR #732, opened 2023-02-22
Add context labeling for toggle buttons

joedolson:
All toggle buttons had the same text "Toggle more information", requiring screen reader users to explore the surrounding context on every button to determine context. Having immediate access to context makes locating information or verifying that you are toggling to correct data significantly faster.


joedolson:
See #713

johnbillion:
This reminds me that I started writing an accessibility statement for my plugins at the end of last year but got a bit stuck writing anything useful. I need to get back to that.

PR #734, opened 2023-02-23
Add a second waiting loop for MariaDB

szepeviktor:
@johnbillion This is so functional.


szepeviktor:
Seems like success.
https://github.com/johnbillion/query-monitor/pull/734/checks#step:14:14
There is a second iteration.

johnbillion:
Kösz! I don't know why this is so difficult.

szepeviktor:
> I don't know why this is so difficult.

Everything is so difficult.
... if you dig deep enough. The IT industry (and others too) are quickly getting worse. The new norm is to know 1 IDE only. Underlying technologies are SCIENTIFIC!

PR #744, opened 2023-03-07
Strict_Type causes issue

debugHere:
The addition (in 3.11.0) of `declare(strict_types = 1);` on line 1, causes an issue on the frontend, throwing
 `Fatal error: Uncaught Error: file_exists() expects parameter 1 to be a valid path, bool given in .../plugins/query-monitor/collectors/languages.php on line 169`

Appears to be expecting a string (file_path) but object/bool given, throwing an error.

### DIFF
--- a/collectors/languages.php
+++ b/collectors/languages.php
@@ -166,7 +166,7 @@ class QM_Collector_Languages extends QM_DataCollector {
 			),
 		) );
 
-		$found = file_exists( $mofile ) ? filesize( $mofile ) : false;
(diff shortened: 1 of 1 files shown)


johnbillion:
@debugHere Could you send me a list of the active plugins on your site please?

The `$mofile` variable comes from the `load_textdomain_mofile` filter and it should only ever be a string. It looks like there may be a plugin or theme on your site that is attempting to load a translation file incorrectly. I'd like to add some extra error handling to Query Monitor in this case to alert you to the problem.

PR #747, opened 2023-03-12
Make settings constants filterable

tillkruss:
Useful for Object Cache Pro.


johnbillion:
Thanks for the PR. Any particular reason for doing this in the outputter instead of the collector?

tillkruss:
I'm indifferent, just placed it close to the constants.

johnbillion:
I just realised these constants don't even have a collector, it's just part of the HTML dispatcher. So ignore my question :D

PR #751, opened 2023-03-21
Check QM_VERSION Constant Before Defining

michael-sumner:
On WordPress VIP platform, the query-monitor plugin is part of the [VIP Go mu-plugins](https://github.com/Automattic/vip-go-mu-plugins/). Therefore, it will throw an error if the query-monitor plugin is also installed and activated in the `/plugins` directory.

This PR fixes the issue by defining the constant only if it is not yet defined.

Screenshot 2023-03-21 at 14 11 38

### DIFF
--- a/query-monitor.php
+++ b/query-monitor.php
@@ -33,7 +33,9 @@ if ( ! defined( 'ABSPATH' ) ) {
 	exit;
 }
 
-define( 'QM_VERSION', '3.12.0' );
+if ( ! defined( 'QM_VERSION' ) ) {
+	define( 'QM_VERSION', '3.12.0' );
+}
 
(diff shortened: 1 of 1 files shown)


michael-sumner:
I have checked all other constant define blocks and they are all ok

johnbillion:
Thanks for the PR. I'm very hesitant to make this change because it's papering over the underlying issue which is that two versions of the same plugin are active. If that's the case then most of the functionality in the plugin will be happening twice, and with the class autoloading you may be loading different versions of class definitions to what are expected.

I think the solution here is to fix the underlying problem which is that you have a plugin active that's also a mu-plugin, and that's never a good idea.

Cheers!

PR #759, opened 2023-04-27
Call previous error handler at end of `QM_Collector_PHP_Errors::error_handler()`

BrianHenryIE:
We have an MU plugin to mute deprecation warnings. With Query Monitor active, its error handler is never run.

The common pattern is to save the return value of `set_error_handler()` to an instance variable, as you have done with `$previous_error_handler`, and then to call it at the end of the new error handler, if it exists. I have added the latter part.

https://www.php.net/manual/en/function.set-error-handler.php#123459

### DIFF
--- a/collectors/php_errors.php
+++ b/collectors/php_errors.php
@@ -280,11 +280,14 @@ class QM_Collector_PHP_Errors extends QM_DataCollector {
 		 * Filters the PHP error handler return value. This can be used to control whether or not the default error
 		 * handler is called after Query Monitor's.
 		 *
+		 * Return true to stop further error handling, false to run the normal error handler.
+		 *
 		 * @since 2.7.0
 		 *
 		 * @param bool $return_value Error handler return value. Default false.
 		 */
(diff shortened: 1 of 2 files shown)


johnbillion:
Thanks for the PR! I'll need to investigate the wider impact of this change.

BrianHenryIE:
I think I need to move the call to the previous error handler to before the filter so it is always run. Currently the filter could short-circuit and return true without calling it. E.g. here is an error handler which logs the error if it's relevant to the plugin, I don't think it's Query Monitor's place to decide _not_ run this code: [bh-wp-logger](https://github.com/BrianHenryIE/bh-wp-logger/blob/7129388bee4c9e49e7dbfc6a1b5eb0ffd9a42751/src/php/class-php-error-handler.php#L91)

I'm having trouble spinning up the test environment (although I obviously had it running already). I'll try again soon.

kagg-design:
This might be fixed by #938.

PR #760, opened 2023-04-28
Prune PHPStan ignores

szepeviktor:
Remove `reportUnmatchedIgnoredErrors: false` and 1 ignored error as the extension can handle `apply_filters`

Plus making regexps more readable.


szepeviktor:
Thanks.
Like a dream come true!

PR #766, opened 2023-05-13
GFM-related finding 😍

szepeviktor:
Not a PR.
A technology demo: https://github.com/szepeviktor/johnbillion_query-monitor/blob/patch-1/readme.txt 😍 

Needs double spaces in headers.


johnbillion:
Thanks for this, but a readme.txt file for a WordPress plugin is a special case, it's not markdown even though it mostly looks like it. I'm not worried about how it renders on GitHub.

PR #768, opened 2023-05-31
Add conditional functions

crstauf:
A few conditional functions that are cool (IMO) but were missing from QM.

`wp_is_site_protected_by_basic_auth()` may be more appropriate for the Environment panel.

### DIFF
--- a/collectors/conditionals.php
+++ b/collectors/conditionals.php
@@ -49,6 +49,8 @@ class QM_Collector_Conditionals extends QM_DataCollector {
 			'is_feed',
 			'is_front_page',
 			'is_home',
+			'is_local_attachment',
+			'is_login',
 			'is_main_network',
 			'is_main_site',
 			'is_month',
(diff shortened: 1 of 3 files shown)


johnbillion:
Thanks! To be honest the only one of these which is really a conditional is `is_login()`, so yes let's add that please. `is_local_attachment()` requires a `$url` parameter.

johnbillion:
Not sure what's up with the test failures, I'll take a look.

crstauf:
Agh, missed that (re: `is_local_attachment()`).

I do really like `is_protected_endpoint()`, and think it could be valuable in some cases. Is there somewhere better for that to go?

crstauf:
Moved `wp_is_site_protected_by_basic_auth()` to Environment panel.

crstauf:
Hmm, I don't recall closing this; will re-open. 🤷‍♂️

PR #773, opened 2023-06-16
Add panel for `wp_mail()` uses

crstauf:
See https://github.com/users/johnbillion/projects/8/views/3?pane=issue&itemId=21798645.


crstauf:
I want to include a _Headers_ column, but it seems the headers array isn't populated in any of the actions (despite being included). 😖

crstauf:
After analyzing `wp_mail()` more, the following headers are removed prior to the actions: `to`, `cc`, `bcc`, `reply-to`; explains why it seemed the `headers` array entry was empty.

crstauf:
2023-06-20T051346
2023-06-20T051439

PR #776, opened 2023-06-20
Fix (at least some) testing errors

crstauf:


crstauf:
@johnbillion Well, this is all I was able to figure out (updating Composer package). We figured out in Slack that PHP 7.2 is not compatible with PHP 9.6.9, but I'm not sure how to fix that.

PR #790, opened 2023-07-10
Fixes misleading redirect shown for HTTP API request

### LINKED ISSUE #738: Misleading redirect shown for HTTP API request
An HTTP API request to the Gravity Forms API with a trailing question mark gets marked as having been redirected on the receiving end because the trailing question mark is stripped somewhere. Need to investigate why this is flagged as a redirect.

crstauf:
Fixes #738.

Change from matching one or more to zero or more.


johnbillion:
Thanks!

PR #791, opened 2023-07-11
Add toggle for HTTP query parameters

crstauf:
See https://github.com/users/johnbillion/projects/8/views/3?filterQuery=&pane=issue&itemId=21799650.

2023-07-11T193003

Query parameters for redirected URL will not show (assumes same as initial URL):

2023-07-11T194032

```php
add_action( 'init', static function () {
	wp_remote_get( 'http://querymonitor.com/?bar=foo' );
	wp_remote_get( 'https://querymonitor.com/?foo=bar&caleb=ryan' );
} );
```


crstauf:
This will conflict with #790.

crstauf:
2023-07-13T160323

PR #805, opened 2023-08-08
Make docker-compose.yml a regular file

szepeviktor:
Symlinks have way more problems than regular files.

- Windows
- editing
- [_please find a third one_](https://docs.docker.com/compose/compose-file/14-include/)

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -23,6 +23,10 @@ You can clone this repo and activate it like a normal WordPress plugin, but you'
 * [Composer](https://getcomposer.org/)
 * [Node](https://nodejs.org/)
 
+To run the tests, you'll also need:
+
+* [Docker Desktop](https://www.docker.com/desktop) running Docker Compose version 2.20 or higher
+
 ### Setup
 
 1. Install the PHP dependencies:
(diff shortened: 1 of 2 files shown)


szepeviktor:
What could happen inside `wpcli`??
May I add `healthcheck` to it?

szepeviktor:
There are relative paths in `docker-compose.yml`.
Where are they? I could not find them.

johnbillion:
Nice one. This might also allow me to remove the `COMPOSE_PROJECT_NAME` usage.

CONTRIBUTING.md
```diff
@@ -23,6 +23,10 @@ You can clone this repo and activate it like a normal WordPress plugin, but you'
 * [Composer](https://getcomposer.org/)
 * [Node](https://nodejs.org/)
 
+To run the tests, you'll also need:
+
+* [Docker Desktop](https://www.docker.com/desktop) running Docker Compose version 2.20 or higher
```
szepeviktor, line 28: What if someone has only the daemon and client?

I feel I am the only ~~weirdo~~ one not using GUI tool.

johnbillion: Telling people to install Docker Desktop is just the easiest way to get set up, especially for users with little or no experience with containers. If you know what you're doing then you probably know you can use a daemon and client instead of Docker Desktop.

szepeviktor: > for users with little or no experience

Yes.
I am totally unable to talk to them about work.

PR #810, opened 2023-09-05
Improve check for Debug Bar existence

### LINKED ISSUE #809: Debug Bar Compatibility: Check if `class_exists( 'Debug_Bar', false )` is not enough in some scenarios
If Jetpack is enabled, it will run `class_exists( 'Debug_Bar' )`, automatically loading the QM version of that class. Then, when QM checks if `class_exists( 'Debug_Bar' )` it will return true (because Jetpack loaded it first) and return early. In that scenario, the Debug Bar compatibility layer will never run.

This is related to https://github.com/10up/debug-bar-elasticpress/issues/77

felipeelia:
Closes #809 

Instead of simply checking if the class exists, this PR changes the code to actually check if the global variable was set and is an instance of the Debug_Bar class.

### DIFF
--- a/collectors/debug_bar.php
+++ b/collectors/debug_bar.php
@@ -66,7 +66,8 @@ function register_qm_collectors_debug_bar() {
 
 	global $debug_bar;
 
-	if ( class_exists( 'Debug_Bar', false ) || qm_debug_bar_being_activated() ) {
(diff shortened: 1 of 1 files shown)


johnbillion:
Thanks, I'll take a look at this at the weekend. Not sure why the acceptance tests have broken.

johnbillion (approved):

PR #824, opened 2023-10-25
null on trim

lazardanlucian:
We have some cases where our pages in wp-admin don't work.

Fatal error: Uncaught Error: trim(): Argument #1 ($string) must be of type string, null given in /var/www/wp-content/mu-plugins/query-monitor/collectors/raw_request.php on line 73

### DIFF
--- a/collectors/raw_request.php
+++ b/collectors/raw_request.php
@@ -70,7 +70,9 @@ class QM_Collector_Raw_Request extends QM_DataCollector {
 		$raw_headers = headers_list();
 		foreach ( $raw_headers as $row ) {
 			list( $key, $value ) = explode( ':', $row, 2 );
-			$headers[ trim( $key ) ] = trim( $value );
+			if ( null !== $key && null !== $value ) {
+				$headers[ trim( $key ) ] = trim( $value );
+			}
 		}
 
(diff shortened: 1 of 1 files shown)


johnbillion:
Thanks for the PR. It sounds like your server is outputting an HTTP header is an unexpected format. Can you provide a list of the headers in the response so we can see what's up?

lazardanlucian:
Hello, I found the issue.
There was a header set without a value, from a php snippets plugin.

This still could be useful here, since a header without a value will not break anything, why have it break qm

lazardanlucian:
array (size=3)
  0 => string 'X-Powered-By: PHP/8.0.27' (length=24)
  1 => string 'Sample-Header-No-Value' (length=22)
  2 => string 'Test: Value' (length=11)
  
  this will break QM and wp-admin pages where Sample header exists

lazardanlucian:
@johnbillion Although the RFC mandate a header value, I'm not sure if the application should break if there is a header without a value.

So our problem is fixed, so it's up to you how you want to treat this 🤖

johnbillion:
Thanks, I'll make some adjustments to the header handling to handle this.

PR #838, opened 2023-11-18
Ignore wp_safe_redirect in redirects trace

ssnepenthe:
Apologies if this was by design but I noticed the redirect trace was stripping the actual wp_redirect call but not wp_safe_redirect


johnbillion:
Thanks!

PR #839, opened 2023-11-22
Fix Fatal error: Object of class WP_Term could not be converted to string

rubas:
## Error
`Fatal error: Uncaught Error: Object of class WP_Term could not be converted to string in /query-monitor/collectors/theme.php on line 119`

## Reproduce

1. Define a taxonomy called `sidebar`
2.  `wp/wp-includes/taxonomy.php:1015` executes `$_term = apply_filters( "get_{$taxonomy}", $_term, $taxonomy );`
3.  `web/app/plugins/query-monitor/collectors/theme.php` catches the action `( 'get_sidebar', array( $this, 'action_get_position' ));`
4. and the function `action_get_position` will explode

### DIFF
--- a/collectors/theme.php
+++ b/collectors/theme.php
@@ -102,6 +102,10 @@ class QM_Collector_Theme extends QM_DataCollector {
 	 * @return void
 	 */
 	public function action_get_position( $name ) {
+	    	if ( !is_string( $name ) && !is_null( $name ) )  {
+      			return;
+    		}
+		
 		$filter = current_filter();
 		$trace = new QM_Backtrace( array(
 			'ignore_hook' => array(


johnbillion:
Yikes. I'm happy to put in a workaround, but I would strongly recommend using a different name for your taxonomy to avoid the problem altogether.

johnbillion:
I've decided to close this off. There are really too many potential problems using a taxonomy with a name that matches an entity in WordPress, and no doubt other plugins that rightfully expect the parameter that's passed to this hook to be a string. I know it's frustrating, it's a shame that there are so many actions and filters in WordPress that have overly generic names.

Cheers!

PR #867, opened 2024-05-08
Prevent an infinite loop when logging doing it wrong calls and deprecated calls

### LINKED ISSUE #840: `_doing_it_wrong()` before `$wp_query` can lead to OOM fatals from new `QM_Collector_Doing_It_Wrong` class
To reproduce:

```
function test_locale_filter( $locale ) {
    $meta_locale = false;

    if ( ! is_attachment() ) {
        $id = get_the_ID();

        if ( $id ) {
            $meta_locale = get_post_meta( $id, 'l10n', true );
        }
    }

    if ( $meta_locale ) {
        return $meta_locale;
    }

    return $locale;
}

add_filter( 'locale', 'test_locale_filter', 20, 1 );
```

You'll see that it OOMs via:
```
PHP message: Fatal error: Allowed memory size of 536870912 bytes exhausted (tried to allocate 20480 bytes) in /wp/wp-content/mu-plugins/query-monitor/classes/Backtrace.php on line 137
```

Commenting out https://github.com/johnbillion/query-monitor/blob/0741b15ea0bc05dc9b6fd71af246cf83cbc45f33/collectors/doing_it_wrong.php#L114 fixes the issue.

johnbillion:
Fixes #840

### DIFF
--- a/collectors/doing_it_wrong.php
+++ b/collectors/doing_it_wrong.php
@@ -16,6 +16,11 @@ class QM_Collector_Doing_It_Wrong extends QM_DataCollector {
 
 	public $id = 'doing_it_wrong';
 
+	/**
+	 * @var bool
+	 */
+	private $collecting = false;
+
 	public function get_storage(): QM_Data {
 		return new QM_Data_Doing_It_Wrong();
 	}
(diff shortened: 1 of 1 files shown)


szepeviktor (approved):

rebeccahum:
Amazing! Will this be included in the next version?

PR #874, opened 2024-07-09
Fix Uncaught TypeError: parse_url(): Argument #1 () must be of type string, null given

rebeccahum:


szepeviktor:
You need to also add a question mark to line 30: `url: string,`

rebeccahum:
Done, thanks!

johnbillion:
Thanks for the PR, I'll take a proper look soon 👍

johnbillion:
Thanks for the PR @rebeccahum but I've decided not to merge this. The `url` element here ultimately comes from the `$url` parameter passed to the `wp_remote_*()` function, which expects a string. That parameter is passed through a bunch of functions and filters, all of which expect it to be a string.

If it's not a string, then that's a developer error, so it's not something that Query Monitor should guard against.

Cheers!

PR #877, opened 2024-07-20
Go on a spacewalk 👨‍🚀

szepeviktor:
@johnbillion This intro is a gift for you 🎁 
https://starwarsintrocreator.kassellabs.io/#!/EO2E7oVWMbEJ5wC8fXSj

These tags would be read in 
https://github.com/johnbillion/query-monitor/blob/2b58634b37b8e7de52d8363c741053f42b7c8079/classes/QueryMonitor.php#L48

```php
$collectors = array_filter(
    get_declared_classes(),
    static function (string $class): bool {
        return str_starts_with($class, 'QM_Collector_');
    }
);
// Run ReflectionClass::getDocComment on $collectors
// Parse @qm-collector tag
// Run add_filter
```
Similar to https://github.com/szepeviktor/SentencePress/blob/412fef6c8d6fed04cfb429cec11f866df4237f11/src/HookAnnotation.php#L43-L53

What do you think?

### DIFF
--- a/collectors/assets_styles.php
+++ b/collectors/assets_styles.php
@@ -10,7 +10,7 @@ if ( ! defined( 'ABSPATH' ) ) {
 }
 
 /**
- * qm-collectors assets_styles
+ * qm-collector assets_styles
  */
 class QM_Collector_Assets_Styles extends QM_Collector_Assets {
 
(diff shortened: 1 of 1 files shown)


johnbillion:
I'm a fan of declarative coding but this is a bit too far from the WordPress way. I've got a list of 100 things that are higher priority too! :D

szepeviktor:
Somehow my brain associates "the WordPress way" with people who write scripts ...

PR #879, opened 2024-07-27
Add panel to display registered assets

crstauf:


johnbillion:
Neat, I like it. Thanks. I think I would be happy to have this panel available without the `QM_SHOW_ALL_ASSETS` constant being required.

crstauf:
Great! Will make that change.

crstauf:
@johnbillion Are you opposed to adding a sort to the _Handle_ column? I'd prefer that to forcing it to be sorted, as there might be a situation where seeing the order that assets are registered is helpful.

crstauf:
Hmmm, forgot sorting was limited to numbers. 🤔

johnbillion:
It should be possible to sort the handle column yeah

crstauf:
@johnbillion I can't add you as a reviewer unfortunately, but please take a peek here: https://github.com/crstauf/query-monitor/pull/1.

johnbillion:
I would like to get this in, but it's triggering a PHP warning on the front end.

```
Trying to access array offset on value of type null	
wp-includes/script-loader.php:2068
wp_style_loader_src()
apply_filters('style_loader_src')
do_action('shutdown')
shutdown_action_hook()
```

The error is inside `wp_style_loader_src()` which attempts to read the `url` property on an admin colour scheme that doesn't exist, presumably because admin scheme colours aren't registered on the front end.

The same problem might occur for other styles or scripts that are only intended for the admin area or only for the front end.

Thoughts?

crstauf:
@johnbillion This seems to only occur because `$_wp_admin_css_colors` isnt' set. Adding the below solves the error:

```php
global $_wp_admin_css_colors;

if ( ! isset( $_wp_admin_css_colors ) || is_null( $_wp_admin_css_colors ) ) {
	$_wp_admin_css_colors = [ 'fresh' => (object) [ 'url' => '' ] ];
}
```

Is that an acceptable solution from your perspective?

johnbillion:
My worry is whether this will result in errors for other third-party scripts and styles that are enqueued in the admin area and don't expect to be processed in the front end, and vice versa. Probably uncommon, but it's hard to tell.

crstauf:
@johnbillion If the asset is globally registered, but only enqueued on one side, all of the necessary data is still available. If an asset is registered only on one side, then it won't be listed.

This error only occurs because there's a dynamic component to `colors` that's handled in `wp_style_loader_src()`, which no themes or plugins would be able to do (without editing core).

Really I think the way WordPress handles this should be changed, but that's a separate conversation.

PR #883, opened 2024-08-07
Add id attribute to qm-title-heading select

arthurshlain:
Fix for Chrome issue: A form field element has neither an id nor a name attribute.

Actually there has no form, but Chrome marks it as issue in Developer console:

image


johnbillion:
Thanks for the PR 👍

PR #901, opened 2024-09-09
check if rows are countable

davidsword:
fatal errors catching this can be null sometimes, add is_countable check to prevent fatals.


johnbillion:
Thanks for the PR David! What can lead to this not being countable?

johnbillion:
@davidsword Were you able to look into this any further? Any info you have could be helpful.

PR #909, opened 2024-10-11
Fix #908

### LINKED ISSUE #908: Visibility of error messages
Query Monitor display the error messages, include stack.

I guess the CSS is too hard to display the message and should be changed, see screenshot.
image

Problem in https://github.com/johnbillion/query-monitor/blob/9b9a31c77ac05820e2c6d61b46672b731ad43d82/assets/_styles.scss#L1475

Remove the position and it should work.
After removal of the line.
image

bueltge:
The removal of this position should be fixed the #908 to display the whole message.

### DIFF
--- a/assets/_styles.scss
+++ b/assets/_styles.scss
@@ -1472,7 +1472,6 @@ body.admin-color-light #wp-admin-bar-query-monitor:not(.qm-all-clear):not(:hover
 	background: #fff !important;
 	max-width: 700px !important;
 	clear: both !important;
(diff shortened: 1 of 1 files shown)


johnbillion:
Thanks for the PR Frank, I'll take a look at this soon 👍

johnbillion:
The reason this uses absolute positioning (with a high z-index) is so that it appears above other elements on the page when the fatal error occurs in an awkward location.

I need to look into this further.

PR #922, opened 2024-12-03
SLSA provenance attestation

johnbillion:
See https://github.com/johnbillion/action-wordpress-plugin-attestation/pull/12

### DIFF
--- a/.github/workflows/deploy-tag.yml
+++ b/.github/workflows/deploy-tag.yml
@@ -43,8 +43,6 @@ jobs:
       id-token: write
     timeout-minutes: 70
     needs: deploy
-    outputs:
-      base64-subjects: ${{ steps.attest.outputs.base64-subjects }}
     steps:
       - name: Attest
         id: attest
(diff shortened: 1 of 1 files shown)


szepeviktor:
You want two things in one workflow ...

johnbillion:
Yeah it's not always easy to determine whether an action is a composite

szepeviktor:
It is a reusable one. It cannot live as a step.

> reusable workflows should be referenced at the top-level `jobs.*.uses' key, not within steps

szepeviktor:
I think this level3 thing must be in your build process.
https://docs.github.com/en/actions/security-for-github-actions/using-artifact-attestations/using-artifact-attestations-and-reusable-workflows-to-achieve-slsa-v1-build-level-3

johnbillion:
I think I rabbit-holed on the actions provided by the SLSA framework due to the isolation requirement for level 3. The annoying thing about using a reusable workflow for a GitHub build provenance attestation is that you then need to verify it using the name of the repo that contains the reusable workflow that created the attestation.

It looks like I need to read into this more.

.github/workflows/deploy-tag.yml
```diff
@@ -43,9 +43,23 @@ jobs:
...
+        uses: johnbillion/action-wordpress-plugin-attestation@slsa
         with:
           version: ${{ needs.deploy.outputs.version }}
           zip-path: ${{ needs.deploy.outputs.zip-path }}
+          dry-run: true
+  slsa:
+    name: Generate SLSA provenance attestation
+    needs: attest
+    permissions:
+      contents: read
```
szepeviktor:
```suggestion
      contents: write
```

johnbillion: Annoying that the workflow needs to be granted write permission even though the condition to run that job isn't satisfied.

PR #939, opened 2024-12-22
Fix E_STRICT is deprecated.

### LINKED ISSUE #921: PHP 8.4: Constant E_STRICT is deprecated
When I run Query Monitor with PHP 8.4, I get a deprecation message.

`PHP Deprecated:  Constant E_STRICT is deprecated in /wp-content/plugins/query-monitor/collectors/environment.php on line 64`

https://github.com/johnbillion/query-monitor/blob/develop/collectors/environment.php#L55-L66

kagg-design:
## Description

With PHP 8.4, QM shows its own deprecation error in the console: "Constant E_STRICT is deprecated in C:\laragon\www\Projects\query-monitor\collectors\environment.php on line 64".

The current PR fixes the issue.


JJJ (approved):
This looks mergable to me 🙌

### Opinion

I don't _love_ switching to maintaining the `E_` values like this, because the original approach of checking `constant()` is more precise, and that is my personal preference. 

But I've also _never_ seen anyone change the values of these constants, so it does seem like a comparison that is simplified as suggested.

johnbillion (approved):

PR #940, opened 2024-12-22
Fix _load_textdomain_just_in_time error.

### LINKED ISSUE #919: _load_textdomain_just_in_time was called incorrectly
Hi,
I have this message in my php error log since the last WP 6.7 update.
To see this error your site must be in a foreign language and you must have debug mode.

Notice: Function _load_textdomain_just_in_time was called <strong>incorrectly</strong>. Translation loading for the <code>query-monitor</code> domain was triggered too early. This is usually an indicator for some code in the plugin or theme running too early. Translations should be loaded at the <code>init</code> action or later. Please see <a href="https://developer.wordpress.org/advanced-administration/debug/debug-wordpress/">Debugging in WordPress</a> for more information. (This message was added in version 6.7.0.) in /Users/xxxx/Local Sites/clone-reve/app/public/wp-includes/functions.php on line 6114

This isn't the only plugin to have this problem.
You must be in debug mode and with a foreign lang site to reproduce this notice.
Could you update your code please ?

Thanks for all.

Have a nice day.

Frédéric

kagg-design:
## Description

Since WP 6.7.0, an annoying message has been displayed in the QM console: "Function _load_textdomain_just_in_time was called incorrectly...".

The proper fix requires much work, such as moving all __() calls after the `after_setup_theme` event. This PR proposes a temporary fix for not outputting noise to QM users.

This PR may fix #919.


johnbillion (approved):
Thanks for the PR 👍

PR #987, opened 2025-05-09
Fix HTML in php_errors

szepeviktor:
There was no space
```html
data-qm-component="Plugin: woocommerce non-core"class="qm-warn">
```


szepeviktor:
Very glad to contribute.

PR #1024, opened 2025-08-22
Update Html.php

mehrazmorshed:
parse_url() is discouraged because of inconsistency in the output across PHP versions; use wp_parse_url() instead.

### DIFF
--- a/dispatchers/Html.php
+++ b/dispatchers/Html.php
@@ -73,7 +73,7 @@ class QM_Dispatcher_Html extends QM_Dispatcher {
 	 * @return bool Should the authentication cookie be secure?
 	 */
 	public static function secure_cookie() {
-		return ( is_ssl() && ( 'https' === parse_url( home_url(), PHP_URL_SCHEME ) ) );
+		return ( is_ssl() && ( 'https' === wp_parse_url( home_url(), PHP_URL_SCHEME ) ) );
 	}
 
 	/**


johnbillion:
Thanks for the PR @mehrazmorshed. Which inconsistencies are you referring to? (Query Monitor supports PHP 7.4 and higher.)

p-windeurope:
From what I can see there were some inconsistencies in the PHP 5.x era (related to schemeless urls). Some new "[inconsistencies](https://www.php.net/manual/en/function.parse-url.php#refsect1-function.parse-url-changelog)" brought by PHP 8.0 - distinguishing between absent and empty query strings and fragments - don't seem to be normalized in [wp_parse_url](https://developer.wordpress.org/reference/functions/wp_parse_url/#source) though.

If someone decides it's a good idea to overwrite (via `add_filter`) the home_url to something like `/a:1` the return values for `PHP_URL_SCHEME` of `parse_url` and `wp_parse_url` would differ (`false` for PHP, `NULL` for WordPress). But that someone would probably have a bigger problem than just this... 🤷

johnbillion:
Thanks for the PR @p-windeurope but I'm going to close this off. I've never managed to get a good answer from anyone, including the WPCS maintainers, to explain what the actual problems with this function are. Whatever they are it does look like they only existed in the PHP 5 era anyway.

Cheers!

p-windeurope:
Good call on closing. As I mentioned, the only difference (PHP 8.2) I was able to produce was:

```php
parse_url("/a:1", PHP_URL_SCHEME);
# returns false

wp_parse_url("/a:1", PHP_URL_SCHEME);
# returns NULL
```

Although `/a:1` can be a valid URI reference, it is not a valid URI (it's [missing a scheme](https://datatracker.ietf.org/doc/html/rfc3986#section-3.1)). I think it's more than fair to expect that `home_url()` returns valid URI. Furthermore the code in questions checks whether the scheme is `https`, it's irrelevant whether the answer is `false` or `NULL` if the scheme is missing.

PR #1026, opened 2025-08-22
Fix PHP 8.5 deprecations

### LINKED ISSUE #1025: Deprecation Notice: Method ReflectionProperty::setAccessible() is deprecated since 8.5
Since PHP 8.1, calling the Reflection*::setAccessible() methods is no longer necessary as reflected properties/methods/etc will always be accessible. However, the method calls are still needed for PHP < 8.1.

As of PHP 8.5, calling the Reflection*::setAccessible() methods is now formally deprecated and will yield a deprecation notice, which will fail test runs. As of PHP 9.0, the setAccessible() method(s) will be removed.

Silencing the deprecation would mean, this would need to be "fixed" again come PHP 9.0, while the current solution should be stable, including for PHP 9.0.

Ref: https://wiki.php.net/rfc/deprecations_php_8_5#extreflection_deprecations

Pross:
Since PHP 8.1, calling the Reflection*::setAccessible() methods is no longer necessary as reflected properties/methods/etc will always be accessible. However, the method calls are still needed for PHP < 8.1.

As of PHP 8.5, calling the Reflection*::setAccessible() methods is now formally deprecated and will yield a deprecation notice, which will fail test runs. As of PHP 9.0, the setAccessible() method(s) will be removed.

Silencing the deprecation would mean, this would need to be "fixed" again come PHP 9.0, while the current solution should be stable, including for PHP 9.0.

Ref: https://wiki.php.net/rfc/deprecations_php_8_5#extreflection_deprecations

Fixes #1025

### DIFF
--- a/.github/workflows/acceptance-tests.yml
+++ b/.github/workflows/acceptance-tests.yml
@@ -54,6 +54,9 @@ jobs:
           - '8.3'
           - '7.4'
         include:
+          # Latest WordPress on PHP 8.5
+          - wp: '6.8'
+            php: '8.5'
           # Latest WordPress on PHP 8.4
           - wp: '6.8'
             php: '8.4'
--- a/.github/workflows/coding-standards.yml
+++ b/.github/workflows/coding-standards.yml
@@ -44,6 +44,7 @@ jobs:
         label:
           - 'PHP'
         php:
+          - '8.5'
           - '8.4'
       fail-fast: false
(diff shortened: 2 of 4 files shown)


johnbillion:
Thanks for the PR!

Pross:
> Thanks for the PR!

Props to composer which is where I stole it ;)

PR #1030, opened 2025-08-25
Improve Opcode statistics

msaggiorato:
This PR adds a bit of information about Opcode cache. We've been dealing at our company with performance issues regarding opcache and it's configuration on some customer's sites, and wanted to contribute back our learnings by allowing people (and hosts most importantly) to be able to identify problems related to this.

APC is legacy, for which we had to scout for information and get basic cache hit. Not sure how much is it actively used, but nothing should break there.

For OPCache, I've added a bunch of information to the overview tab. Mainly the cache hits for the current request (opcache provides statistics for the whole lifecycle of the FPM process), and also several memory/limits indicators.

Tried to keep UX similar to the rest of QM, hopefully not disturbing anything.

There is a change in function names (due to untying object cache stats from opcode cache stats). I opted for the small refactor in order to keep things clean.

### DIFF
--- a/collectors/object-cache.php
+++ b/collectors/object-cache.php
@@ -12,9 +12,9 @@ if ( ! defined( 'ABSPATH' ) ) {
 /**
  * @extends QM_DataCollector<QM_Data_Cache>
  */
-class QM_Collector_Cache extends QM_DataCollector {
+class QM_Collector_Object_Cache extends QM_DataCollector {
 
-	public $id = 'cache';
+	public $id = 'object-cache';
 
 	public function get_storage(): QM_Data {
 		return new QM_Data_Cache();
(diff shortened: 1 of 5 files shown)


johnbillion:
Thanks for the PR, I'll take a look soon. Don't worry about the failing tests, there's an issue with the PHP 8.3 and 8.4 containers at the moment.

msaggiorato:
Added something back from our prototype tool, to allow to see server limits, even if the statistics cannot be fetched.

msaggiorato:
Adding some screenshots for reference

Regular display
image

When no stats are avaliable
image

Opcache installed, but not enabled
image

And there's a similar text as Object Cache, which is displayed when there's no extension installed, suggesting to ask the host to enable

msaggiorato:
Hey @johnbillion , I know that work on v4 is undergoing. Unsure if you ever got to check this, but would it be a good thing to get back to this once the rendering engine is wrapped for v4?

johnbillion:
Yeah I'd like to get these changes in but I think it doesn't make sense to do it right now seeing as the rendering switch is imminent. 👍 

Let's keep the PR open and I'll decide how best to tackle currently open PRs once 4.0 is out.

PR #1052, opened 2025-10-22
Fix fatal error WordPress v6.9 beta1

jilani53:
A fatal error occurred while using WordPress Directory version 3.20.0, which has now been fixed.
Error screenshot - https://prnt.sc/RvCJ_r3KcrZJ

### DIFF
--- a/.github/workflows/acceptance-tests.yml
+++ b/.github/workflows/acceptance-tests.yml
@@ -46,6 +46,7 @@ jobs:
       matrix:
         wp:
           # Three most recent versions of WordPress
+          - '6.9-beta1'
           - '6.8'
           - '6.7'
           - '6.6'
--- a/.github/workflows/integration-tests.yml
+++ b/.github/workflows/integration-tests.yml
@@ -47,6 +47,7 @@ jobs:
       matrix:
         wp:
           # Three most recent versions of WordPress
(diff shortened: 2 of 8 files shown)


johnbillion:
Thanks for the PR @jilani53!

As you can see from the static analysis errors, it should never be possible for a script module to not include dependencies.

* Has something changed in 6.9?
* Can you provide example `wp_enqueue_script_module()` or `wp_register_script_module()` code from your site that is causing this problem?

Thanks!

desrosj:
Just wanted to note that I also noticed this on my site today. I run the nightly build, and currently on `6.9-beta1-61042`. Nothing changed between yesterday and today except the the changes made leading up to beta 1 being included in the latest nightly.

johnbillion:
Thanks. I can reproduce this on my test site too, it was introduced in https://core.trac.wordpress.org/ticket/63486 . I'll take a proper look.

johnbillion:
Thanks both! This is fixed and I'll get a release out shortly.

johnbillion:
The change in 6.9 beta 1 that caused this has been reverted in beta 2 so it's no longer an issue. 

We've got https://core.trac.wordpress.org/ticket/60597 for exposing a proper public API for enqueued modules.

PR #1058, opened 2025-11-07
Remove calls to deprecated jQuery .focus() method

### LINKED ISSUE #1057: JQMIGRATE: jQuery.fn.focus() event shorthand is deprecated
The script `query-monitor.js` calls the jQuery `.focus()` method, which is deprecated.  This results in the following message being written to the browser console (you may need `SCRIPT_DEBUG` defined as `true` in your `wp-config.php` file):

```
JQMIGRATE: jQuery.fn.focus() event shorthand is deprecated
```

According to the [jQuery `.focus()` documentation](https://api.jquery.com/focus-shorthand/), the fix is simple:

> Instead of `.focus()`, use [`.trigger( "focus" )`](https://api.jquery.com/focus/#trigger2).

siliconforks:
Fixes #1057.

### DIFF
--- a/assets/query-monitor.js
+++ b/assets/query-monitor.js
@@ -119,7 +119,7 @@ if ( window.jQuery ) {
 			}
 
 			show_panel( href );
-			$(href).focus();
+			$(href).trigger('focus');
 			$('#wp-admin-bar-query-monitor').removeClass('hover');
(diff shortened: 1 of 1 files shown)


johnbillion:
Thanks for the PR!

PR #1063, opened 2026-01-10
Improve accessibility of toggle buttons with contextual aria-labels

Ankita19Rathore:
### What this PR does

This pull request improves the accessibility of the "Toggle more information"
buttons used in the HTML output by making their aria-labels contextual and
descriptive.

---

### What was happening before

Previously, all toggle buttons were created using the following aria-label:

aria-label="Toggle more information"

Because this same label was used everywhere, screen readers announced
multiple buttons with identical text. This made it difficult for users
to understand what each toggle button would expand or control.

---

### What was changed

The `build_toggler()` method was updated to accept an optional `$context`
parameter.

**Before:**
```php
protected static function build_toggler() {
	$out = '<button class="qm-toggle" ... aria-label="Toggle more information">...</button>';
	return $out;
}
```

**After:**
```php
protected static function build_toggler( $context = '' ) {
	$label = __( 'Toggle more information', 'query-monitor' );

	if ( $context ) {
		$label = sprintf(
			__( 'Toggle more information about %s', 'query-monitor' ),
			$context
		);
	}

	return '<button class="qm-toggle" ... aria-label="' . esc_attr( $label ) . '">...</button>';
}
```

### Behavior after this change

If context is provided, the aria-label becomes descriptive  
(e.g. `Toggle more information about Jetpack hook callback`).

If no context is provided, the original behavior is preserved.

---

### Why this change is important

Screen reader users rely on aria-labels to understand interactive elements.  
Using repeated generic labels reduces accessibility and usability.

By introducing contextual aria-labels, each toggle button can now be
clearly distinguished, aligning with accessibility best practices.

---

### Scope and safety of change

- The change is limited to the `build_toggler()` method only.
- No visual or functional behavior has been altered.
- Existing functionality remains backward-compatible.
- No unrelated code was modified.

---

### DIFF
--- a/output/Html.php
+++ b/output/Html.php
@@ -391,25 +391,21 @@ abstract class QM_Output_Html extends QM_Output {
 	 * @return string Markup for the toggle control.
 	 */
 	protected static function build_toggler( string $context = '' ) {
-		$label = __( 'Toggle more information', 'query-monitor' );
-
-		if ( $context ) {
-			$label = sprintf(
-				/* translators: %s: Context for the toggle button. */
+		$label = $context
+			? sprintf(
+				/* translators: %s: Context for the toggle button, e.g. a function name. */
 				__( 'Toggle more information about %s', 'query-monitor' ),
 				$context
-			);
-		}
+			)
+			: __( 'Toggle more information', 'query-monitor' );
 
-		$out = '<button class="qm-toggle" data-on="+" data-off="-" aria-expanded="false" aria-label="' .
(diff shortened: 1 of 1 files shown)


Ankita19Rathore:
I have fixed the issue mentioned above.
Please review the changes.

johnbillion:
Thanks for the PR @Ankita19Rathore . I'll take a look soon.

Ankita19Rathore:
Thanks for the review!

PR #1064, opened 2026-01-16
Prevent fatal error when request value is null

vanshika2408:
This PR prevents a PHP fatal error by ensuring the request value is a string before passing it to preg_match().

Fixes a case where a null value could be encountered, particularly on PHP 8+.

### DIFF
--- a/collectors/request.php
+++ b/collectors/request.php
@@ -310,9 +310,12 @@ class QM_Collector_Request extends QM_DataCollector {
 		$rewrite_rules = $wp_rewrite->rules;
 
 		foreach ( $rewrite_rules as $match => $query ) {
-			if ( preg_match( "#^{$match}#", $this->data->request['request'] ) ) {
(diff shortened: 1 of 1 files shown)


johnbillion:
Thanks for the PR @vanshika2408 ! How can this be triggered please?

vanshika2408:
> Thanks for the PR @vanshika2408 ! How can this be triggered please?

Thanks for checking this!

I wasn’t able to pin this down to a single, clean repro case, but while looking at the code it seems possible for $this->data->request['request'] to be null in some edge cases, for example very early in the request lifecycle, or around redirects (e.g. right after login) where the request data isn’t fully populated yet.

On PHP 8+, that ends up causing a fatal error when it’s passed straight into preg_match().

The change here is just a defensive guard to make sure we only call preg_match() with a string, without changing behavior when the value is present.

Issue #1069, opened 2026-02-06
Feature Request: Add a diff for added/removed queries between page loads of the same page


leonidasmi:
During my usual workflow, I regularly refresh the same page in my test site to understand whether changes I've made in my own plug-in have resulted in performance regressions. The way I do that is by trying to understand whether queries have been added between page loads. 

A nice visual aid for the above would be to have a separate tab that has a diff of queries when a page gets refreshed. Something like the below:

<img width="1063" height="384" alt="Image" src="https://github.com/user-attachments/assets/44f0b9b7-e6ca-4f61-a16d-61726e06ed10" />

---

I have created [a POC](https://github.com/johnbillion/query-monitor/pull/1070), that implements a rough version of this and I'm sharing it here in case it's useful. If it's in the right direction and the feature request makes sense, I'm willing to turn it in a proper PR at some point (of course some initial feedback would be greatly appreciated). If not, feel free to treat it as the POC that it is :) 

TIA!

Issue #1068, opened 2026-02-04
Query Monitor 4


johnbillion:
This issue is for general tracking and information about the next major release of Query Monitor, version 4, which switches from generating HTML output server-side in PHP to rendering components client-side with Preact.

## Testing

You can download a built pre-release version of Query Monitor 4 from [one of the 4.x releases here](https://github.com/johnbillion/query-monitor/releases) and then install and activate it just like a normal plugin.

## Features

While there are some visual tweaks, _this should be a mostly invisible change to end users_. The big underlying improvement is the rendering performance and enablement of Fantastic Future Features.

This change radically improves rendering performance as it removes the need to output a huge blob of HTML server-side. Individual panels are rendered on-demand client side using data from a relatively performant data object and a fixed-size JavaScript bundle.

This change also facilitates future enhancements such as displaying full data from other requests (historical requests, Ajax/REST requests), lazily loading data, showing client-side collected data, remixing data into different views (graphs, timelines), and generally allowing the display of data to be disconnected from the current request.

With this change QM no longer has any external JavaScript dependencies, including jQuery, React, wp.i18n, etc. Everything is self-contained in the 100KB bundle.

## Development

Development is now happening in the default `develop` branch. No further bug fixes or enhancements will be accepted into the 3.x version.

## Back-compat

- Back-compat support for server-side PHP rendered panels in third-party plugins is in place. Such panels are indistinguishable from the client-side rendered panels that they sit alongside.
- In the initial release there will be no facility for a third-party plugin to implement its own client-side rendered panels. This is a future enhancement.

## To-dos

There are still some to-dos to be to-done before the stable release. [You can see the project board here](https://github.com/users/johnbillion/projects/9).

## Bug reports

Are you testing Query Monitor 4? You're amazing. Bugs can be reported by opening a regular issue here on the GitHub repo.

PR #1070, opened 2026-02-06
Create new Query Diff tab

leonidasmi:
> [!NOTE]  
> Everything below has been copied over (after resolving conflicts and aligning with the brand new `develop` branch) to [this PR](https://github.com/johnbillion/query-monitor/pull/1123).

**Disclaimer**: This is a POC and not a finished product. The code has been briefly validated, but may very well lack proper security, privacy or performance measures.

## Context

* This is regarding [this feature request](https://github.com/johnbillion/query-monitor/issues/1069)
  * Adds a query diff tab under the database queries tab.
  * The new query diff tab is populated by a diff of queries between page loads whenever we refresh a page. 
  * That way, developers have an easy visual to understand whether their current work has resulted in performing regressions or improvements.

## Relevant technical choices:

* I went with storing the query signatures in session storage instead of using the database, mostly because of privacy concerns in sites with multiple users.
* The feature is getting enabled only when the user opens the new Query Diff panel. Two notes:
  * We did this to reduce the impact by avoiding storing the queries snapshot on page load. If we think doing that is fine, we can add a side effect in the DOMContentLoaded handler that populates the snapshot.
  * We can improve this by:
    * Make opening the Database Queries panel the trigger for this
    * Make it clearer to the user of this, to avoid confusion
* Instead of giving queries_diff it's own panel menu, to avoid creating a new collector I added the new feature as a child of the DB Queries panel menu. That way, I just re-use the `db_queries` collector.

## How to use:
* **Enable**: Go to QM Settings → Query Diff → check "Enable query diff tracking"
* **Trigger storage**: Go to the Query Diff panel
* **First load**: Queries are captured and stored in sessionStorage
* **Refresh**: Current queries are compared against stored queries
* **Diff panel**: Appears below the Database Queries table showing:
  * Summary (previous count → current count)
  * Added queries (with SQL and caller)
  * Removed queries (with SQL and caller)

image

## Possible next steps:
* Consider having a more flexible query comparison, by replacing values with placeholders
  * Could make the implementation more hardened for storing sensitive data
  * Would also reduce the noise (false positives) when queries are largely the same (eg. we shouldnt care if the ```UPDATE `wp_usermeta` SET `meta_value` = '1770383389' WHERE `user_id` = 1 AND `meta_key` = 'wc_last_active'``` query is repeated with different meta values
* Add time expiration in the session storage
* Verify the logic of calculating and diff-ing query snapshots - there might be performance chances there (and edge-case bugs)

### Toggling the new setting
* Go to a frontend page and open the QM panels
* Confirm that you dont see in the local storage of your browser (via the Application tab in dev tools), that there's no `qm-query-diff-enabled` entry.
* Enable the `Enable query diff tracking` setting
  * Confirm that you now see a `qm-query-diff-enabled` entry in local storage and it is true.
  * Open the Query Diff panel and confirm that you now see the `qm-query-diff-data` entry in the session storage 
    * _NOTE: we can improve this, so that we can make opening the Database Queries panel the trigger to fill in the session storage_
  * Confirm that you see the `qm-query-diff-data` entry in the session storage data via the Application tab in dev tools
* Go to the settings and disable the `Enable query diff tracking` setting
  * Confirm that you see no `qm-query-diff-data` entries anymore in the session storage data
  * Also confirm that you no longer see the `qm-query-diff-enabled` entry as true in your local storage. It should be false now.

### Moving to a different page:
* While having the setting enabled, visit the Query Diff panel and confirm that the `qm-query-diff-data` entry in the session storage has a URL attribute that's equal to the current page's URL
  * Go to a different page and once loaded, confirm that the that the `qm-query-diff-data` entry in the session storage has been updates so the URL attribute is now equal to the new page's URL
  * Confirm that you get a `Refresh this page to compare queries between page loads` message in the Query Diff panel.


johnbillion:
This looks interesting!

I will say that this is probably best done properly [once version 4 is out](https://github.com/johnbillion/query-monitor/issues/1068) and we've switched to rendering the panels with Preact and gotten rid of the jQuery soup. But I like the idea a lot.

leonidasmi:
Glad you like it!

I'll come back to it once v4 is out, good luck with that.

leonidasmi:
By the way, I'd love to take a stab at [it](https://github.com/johnbillion/query-monitor/pull/1070#issuecomment-3861797543) now, even if 4.0.0 isn't out, if you think it's not a terribly bad idea.

For that, I can branch out of release-4.0.0-alpha.2 and (slowly) port what I have here into the new rendering logic.

johnbillion:
It's up to you. The branch is fairly stable now (although I still need to make changes to the way stack traces are recorded) but I've not written much documentation yet. Data collection hasn't really changed but the CONTRIBUTING.md file details how the data classes are now generated and passed from php to react.

leonidasmi:
@johnbillion was this an auto close or it was a conscious decision?

If the former, I'm still interested in picking this back up again at some point, since you also thought it'd be a nice addition.

johnbillion:
Looks like it was closed because the target branch was merged. Sorry, wasn't intentional!

Yes more than happy if you'd like to pick this up again, feel free to open another PR.

PR #1093, opened 2026-04-09
Allow non-UTF8 (ie binary) in json

Xilonz:
Fixes situations where raw binary is used in database-queries (ie for efficiently storing IP adresses) 

Previously the encode would fail, causing the panel to not render at all.

### DIFF
--- a/classes/Util.php
+++ b/classes/Util.php
@@ -645,7 +645,7 @@ class QM_Util {
 	/**
 	 * Helper function for JSON encoding data and formatting it in a consistent manner.
 	 *
-	 * @deprecated Use json_encode() directly with the appropriate options instead.
+	 * @deprecated Use wp_json_encode() directly with the appropriate options instead.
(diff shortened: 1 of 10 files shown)


johnbillion:
Thanks for the PR @Xilonz. Do you have some example code to test this with please?

Xilonz:
@johnbillion I didn't open an issue for this, if you need/want a more detailed explanation or think this isn't enough to prevent the issue I'm happy to do so.

We use MalCare security; which stores the IP adress in binary format. 

Anything that is not utf-8 encoded would've caused the plugin to fail, so there are probably other scenario's the json_encode will fail as well.

johnbillion:
Thanks again @Xilonz. I decided to switch everything back to using `wp_json_encode()` which recursively converts non-UTF8 characters to UTF8, and generally handles invalid UTF8 sequences.

Issue #1087, opened 2026-04-08
Query Monitor 4.0.0 and 4.0.1 output raw data on frontend for logged-in admin, while 3.20.4 works normally


raygorgor:
After upgrading Query Monitor from 3.20.4 to 4.0.0 and 4.0.1, the frontend starts displaying garbled/raw Query Monitor data for logged-in admin users.

What I tested:
- Query Monitor 3.20.4: works normally
- Query Monitor 4.0.0: frontend output breaks / raw data appears
- Query Monitor 4.0.1: same problem as 4.0.0
- Disabling Query Monitor restores the frontend immediately
- Downgrading back to 3.20.4 restores normal behavior

Environment:
- WordPress 6.9.4
- WooCommerce
- Blocksy Companion Pro
- FlyingPress
- Rank Math SEO
- WPCodeBox 2
- Meta Box

Notes:
- Public visitors are not affected
- The issue appears only when Query Monitor 4.0.0 or 4.0.1 is active
- I first noticed it after updating Query Monitor, then editing some footer heading tags in the Customizer
- However, the heading change itself does not seem to be the cause, because reverting Query Monitor to 3.20.4 fixes the issue immediately

The raw frontend output included callback / hook-related data such as:
- callback_type
- name
- file
- component
- priority

This may be related to the new client-side rendering introduced in Query Monitor 4.x.

I can provide screenshots if needed.

<img width="920" height="680" alt="Image" src="https://github.com/user-attachments/assets/2e8b68da-32cd-4c5c-8a94-fd5a68e6643f" />

johnbillion:
Thanks for the report. I suspect this is due to some incorrect HTML at the end of your theme.

- Which theme are you using? 
- Does the problem go away if you switch to a different theme? 
- If not, can you try deactivating other plugins on your site one by one to see if any of them are causing a problem?

raygorgor:
Hi John,

Thanks for your reply.

My website is giftsdepot.my, and I’m using the Blocksy theme.

So far, what I’ve observed is:

   - Version 3.20.4 works normally on my site.
   - The issue appears with 4.0.0 / 4.0.1, where raw Query Monitor output
   is shown on the frontend for logged-in admin users， as shown in the photos
   I attached earlier.
   - I have currently removed 4.0.1 and reverted to 3.20.4, and the problem
   disappears under the same site conditions.

I have not yet completed full isolation testing with a different theme or
by deactivating plugins one by one, because this is a live site and it
[...]

PR #1098, opened 2026-04-10
Fixes the issue raw html in frontend by adding JSON_HEX_TAG

tanelaavistu:
Fixes the issue where raw html was sent to the footer of the page (Ref. #1087 )

For us the issue was this block:
```
<!-- wp:html -->
<script src="https://player.vimeo.com/api/player.js"></script>
<!-- /wp:html -->
```

HTML:
`<script src="https://player.vimeo.com/api/player.js"></script>`


johnbillion:
Excellent spot. Thank you.

PR #1123, opened 2026-06-19
Create new Query Diff tab

leonidasmi:
## Summary

* This is regarding [this feature request](https://github.com/johnbillion/query-monitor/issues/1069)
  * Adds a query diff tab under the database queries tab.
  * The new query diff tab is populated by a diff of queries between page loads whenever we refresh a page. 
  * That way, developers have an easy visual to understand whether their current work has resulted in performing regressions or improvements.
  * This feature is disabled for the extension, for keeping things simple for now. Making the feature work there requires quite the overhead, if I'm not mistaken.
  * This PR corresponds to https://github.com/johnbillion/query-monitor/pull/1070 that was automatically closed because its target branch was deleted. I thought it's easier to start fresh, so this PR should be the end result eventually
* [An extra Playwright configuration was added](https://github.com/johnbillion/query-monitor/pull/1123/changes#diff-d32ab54c10db37c1914f4054a214ed9e1171a49361334600822038ee1802b345), to enable JS unit testing
  * This was done mostly to easily test the more complex parts of the PR (the diff-ing algo) in an extensive way
  * As a result, I had to also touch `composer.json`, `CONTRIBUTING.md` and `package.json`, apologies if that's out of line

## Relevant technical choices:

* I went with storing the query signatures in session storage instead of using the database, mostly because of privacy concerns in sites with multiple users.
* The feature is getting enabled only when the user opens the new Query Diff panel. Two notes:
  * We did this to reduce the impact by avoiding storing the queries snapshot on page load. 
  * I've added a relevant `@TODO` in case we want to improve that, possibly by making opening the Database Queries panel the trigger for this

## How to use:
* **Enable**: Go to QM Settings → Query Diff → check "Enable query diff tracking"
* **Trigger storage**: Go to the Query Diff panel
* **First load**: Queries are captured and stored in sessionStorage
* **Refresh**: Current queries are compared against stored queries
* **Diff panel**: Appears below the Database Queries table showing:
  * Summary (previous count → current count)
  * Added queries (with SQL and caller)
  * Removed queries (with SQL and caller)
## Testing

- [x] I have tested this change on a real installation of WordPress (required)

### The main functionality
* Follow the `How to use` steps to verify the main functionality of our feature. Examples on how to do that:
* Be on the homepage and refresh without changing much else in your site. 
  * You should see a summary in the Query Diff panel that says: `No query changes detected between page loads`. 

image

_if you didnt change anything but still saw changes in the queries, this might have been a result of semi-regular operations, like transients expiring, crons firing, or stuff of that nature_
* Be on the homepage, delete the transients of your site (WP CLI: `wp transient delete --all`) and refresh. 
  * You should see a summary in the Query Diff panel that says: `Previous: 29 queries. Current: 34 queries` (actual numbers depend on your site) and below that, a list of transient-related queries that were added since the last page load

image
 
* If you refresh the homepage once again, those transient-related queries will not run again
  * As a result, you should see a summary in the Query Diff panel that says: `Previous: 34 queries. Current: 29 queries` (the numbers from the previous test but swapped ) and below that, a list of transient-related queries that were removed since the last page load

image

* Of course, there are all sorts of way you can verify the results of the query diffing (like adding/removing a mu-plugin that runs a query in each page load and refreshing a page accordingly), the transient deletion is one example of that.

### Edge-case test (diff-ing duplicate queries)
* In order to test whether duplicate queries are properly calculated in the diffs, you can add the following mu-plugin after loading a page:
```
<?php
/**
 * Plugin Name: QM Query Diff Test
 * Description: Runs the same query twice on every page load (at init and at shutdown) for testing the Query Monitor query-diff feature.
 */

function qm_diff_test_run_query() {
	global $wpdb;

	$wpdb->query( "SELECT ID, post_title FROM {$wpdb->posts} WHERE post_status = 'publish' ORDER BY ID ASC LIMIT 5" );
}

add_action( 'init', 'qm_diff_test_run_query' );

// QM renders its output on shutdown at priority 9, so run earlier than that or the query won't be captured.
add_action( 'shutdown', 'qm_diff_test_run_query', -10 );
```
* Now refresh the page and confirm that you see the Query Diff panel reporting at least two added queries, the above duplicated one.
* Remove the mu-plugin, refresh the page and confirm that you see the Query Diff panel reporting at least two removed queries, the above duplicated one.

### Toggling the new setting
* Go to a frontend page and open the QM panels
* Confirm that you dont see in the local storage of your browser (via the Application tab in dev tools), that there's no `qm-query-diff-enabled` entry (or that it's false, if you have ever enabled it before)
* Enable the `Enable query diff tracking` setting
  * Confirm that you now see a `qm-query-diff-enabled` entry in local storage and it is true.
  * Open the Query Diff panel and confirm that you now see the `qm-query-diff-data` entry in the session storage. That entry should contain
    * The URL we're on
    * The timestamp we accessed the page
    * The list of queries ran during the page load
* Go to the settings and disable the `Enable query diff tracking` setting
  * Confirm that you see no `qm-query-diff-data` entries anymore in the session storage data
  * Also confirm that you no longer see the `qm-query-diff-enabled` entry as true in your local storage. It should be false now.

### Moving to a different page:
* While having the setting enabled, visit the Query Diff panel and confirm that the `qm-query-diff-data` entry in the session storage has a URL attribute that's equal to the current page's URL
  * Go to a different page and once loaded, confirm that the that the `qm-query-diff-data` entry in the session storage has been updates so the URL attribute is now equal to the new page's URL
  * Confirm that you get a `Refresh this page to compare queries between page loads` message in the Query Diff panel.

### Issues with session storage
* If session storage is unavailable (for example when session storage quota is exceeded), we disable the feature with a warning
* Load a page, open the Query Diff tab, move away from that tab and refresh the page
* in the console run: `Object.defineProperty( window, 'sessionStorage', { get() { throw new Error( 'blocked' ); } } );`
* This mimics what browsers do when storage is blocked — accessing window.sessionStorage itself throws.
* Now open the Query Diff panel: you will see a warning:
image

* Clean up: just reload (the override doesn't survive navigation).
* Now for issues when cleaning up the session storage data, go to the QM settings 
* Run again in the console: `Object.defineProperty( window, 'sessionStorage', { get() { throw new Error( 'blocked' ); } } );`
* Deactivate the Query Diff feature
* Confirm that you see the warning:
image

### When SAVEQUERIES is false
* Add `define( 'SAVEQUERIES', false );` to your wp-config.php
* Enable the Query Diff feature and refresh a page while expecting queries to be added/removed (eg. by deleting all transients between page loads)
* You will see the `Database queries were not logged, so query diffs cannot be tracked between page loads.` message in the Query Diff tab.

### Extension mode
* We have disabled the feature for the extension for now
* Use the extension and go to the Query Monitor tab in DevTools
* Confirm that:
  * You dont see the Query Diff section in the settings
  * You dont see the Query Diff panel in the menu under Database Queries

## Possible next steps:
* Consider having a more flexible query comparison, by replacing values with placeholders
  * Could make the implementation more hardened for storing sensitive data
  * Would also reduce the noise (false positives) when queries are largely the same (eg. we shouldnt care if the ```UPDATE `wp_usermeta` SET `meta_value` = '1770383389' WHERE `user_id` = 1 AND `meta_key` = 'wc_last_active'``` query is repeated with different meta values
* Add time expiration in the session storage

## AI assistance

- [x] I have used AI assistance to produce this pull request
- [ ] I am an AI agent acting on behalf of a human

## Screenshots

- [x] I have included a before and after screenshot (required, unless there is genuinely no visual change)

### Before
image
image

### After
image
image


johnbillion:
Thanks for the PR @leonidasmi , I'll find some time to look at it soon.

leonidasmi:
thank you @johnbillion, the PR is now at a review-able state, so feel free! I'm interested to see what you think :)

PR #1127, opened 2026-06-25
Guard QM_Timer against reading an unstopped timer

### LINKED ISSUE #1118: Array offest issue in Timer.php
Hello!

My error log is being filled with `PHP Warning:  Trying to access array offset on value of type null in server/htdocs/wp-content/plugins/query-monitor/classes/Timer.php on line 143`. It seems like there is something going on with how QM_Timer's `$end` and/or `$start` variables are being set.

Let me know what other information I can provide to help with troubleshooting this.

wj-yuta-imai:
Guards `QM_Timer::get_time()` and `get_memory()` against being read before the timer is stopped, fixing the `Trying to access array offset on value of type null` warning in #1118.

`$end` is null until `stop()` runs, so reading a started-but-never-stopped timer triggered the warning. This can happen via a `qm/start` without a matching `qm/stop`, and also internally when a block timer is started in `block_editor.php` but `render_block` never fires for that block. Both delta methods now return `0` when the timer hasn't stopped; the stopped path is unchanged.

`get_end_time()` / `get_end_memory()` are intentionally left as-is: they return absolute values, so a `0` sentinel would mislead rather than help.

AI-assisted: prepared with the help of an AI coding agent (Claude) and reviewed before submitting.

Fixes #1118

### DIFF
--- a/classes/Timer.php
+++ b/classes/Timer.php
@@ -140,6 +140,9 @@ class QM_Timer {
 	 * @return float
 	 */
 	public function get_time() {
+		if ( ! is_array( $this->end ) ) {
+			return 0.0;
+		}
 		return $this->end['time'] - $this->start['time'];
 	}
 
@@ -147,6 +150,9 @@ class QM_Timer {
 	 * @return int
 	 */
 	public function get_memory() {
+		if ( ! is_array( $this->end ) ) {
+			return 0;
+		}
 		return $this->end['memory'] - $this->start['memory'];
 	}
 


johnbillion:
Thanks for the PR @wj-yuta-imai . What code did you use to reproduce this problem on your test environment please?

wj-yuta-imai:
Honest answer: I didn't reproduce this on a live site — I diagnosed it from the #1118 stack trace (`Timer.php:143`) and the code.

The mechanism is that `get_time()`/`get_memory()` read `$this->end`, which is `null` until `stop()`:

```php
$t = new QM_Timer();
$t->start();
$t->get_time();   // Warning: Trying to access array offset on null — Timer.php:143
```
(Confirmed on PHP 8.3.)

I haven't pinned the exact caller behind #1118 — QM's own collectors look guarded — so this is really a defensive hardening of `QM_Timer`, returning `0` instead of warning for any caller that reads an unstopped timer. Happy to add a regression test, or to drop it if you'd prefer to keep the class strict and fix callers instead.

